PR #6, opened 2022-10-10
Reset the TaskChain if one of them is cancelled.

klvnraju:
We saw an indication in telemetry that a task in TaskChain gets cancelled and making subsequent queueing of tasks fail.
This change will reset the TaskChain in any task fails or errors out.

### DIFF
--- a/src/cs/Ssh/TaskChain.cs
+++ b/src/cs/Ssh/TaskChain.cs
@@ -49,6 +49,16 @@ internal class TaskChain : IDisposable
 		try
 		{
 			await semaphore.WaitAsync(cancellation).ConfigureAwait(false);
+			
+			if (runInSequenceTask != null && 
+				(runInSequenceTask.IsCanceled || runInSequenceTask.Exception != null))
+			{
+				// If one task in the sequence is cancelled we have to reset runInSequenceTask 
+				// so that all subeequent queueing will succeed.
+				runInSequenceTask = null;
+			}
+
+			cancellation.ThrowIfCancellationRequested();
 			if (runInSequenceTask == null)
 			{
 				runInSequenceTask = Task.Run(


src/cs/Ssh/TaskChain.cs
```diff
@@ -49,6 +49,16 @@ public async Task RunInSequence(Func<Task> task, Action<Exception> onError, Func
...
+			
+			if (runInSequenceTask != null && 
+				(runInSequenceTask.IsCanceled || runInSequenceTask.Exception != null))
+			{
+				// If one task in the sequence is cancelled we have to reset runInSequenceTask 
+				// so that all subeequent queueing will succeed.
+				runInSequenceTask = null;
+			}
+
+			cancellation.ThrowIfCancellationRequested();
```
IlyaBiryukov, line 61: nit: no need because `semaphore.WaitAsync` should have thrown already if `cancellation` was cancelled.

src/cs/Ssh/TaskChain.cs
```diff
@@ -49,6 +49,16 @@ public async Task RunInSequence(Func<Task> task, Action<Exception> onError, Func
 		try
 		{
 			await semaphore.WaitAsync(cancellation).ConfigureAwait(false);
+			
+			if (runInSequenceTask != null && 
+				(runInSequenceTask.IsCanceled || runInSequenceTask.Exception != null))
+			{
+				// If one task in the sequence is cancelled we have to reset runInSequenceTask 
+				// so that all subeequent queueing will succeed.
```
klvnraju, line 57: typo `subsequent`

src/cs/Ssh/TaskChain.cs
```diff
@@ -49,6 +49,16 @@ public async Task RunInSequence(Func<Task> task, Action<Exception> onError, Func
 		try
 		{
 			await semaphore.WaitAsync(cancellation).ConfigureAwait(false);
+			
+			if (runInSequenceTask != null &&
```
IlyaBiryukov, line 53: I think I see another issue. If `preTask` fails on line 112, then `completionSource` is never set, and the task chain will be blocked waiting for it forever. 

Not blocking his PR.

IlyaBiryukov (dismissed):

src/cs/Ssh/TaskChain.cs
klvnraju: `completionSource.SetResult` is done in finally block (line 122) right? so, should get executed even if `preTask` fails.

IlyaBiryukov: Right. I missed it because GH hidden it and I didn't expand

jasongin (approved):

jasongin (approved):

klvnraju:
Ran the tests on my mac machine and results on par with old SSH repo

PR #7, opened 2022-10-18
Get remote ip address as a property for session

klvnraju:
Required for telemetry https://github.com/microsoft/basis-planning/issues/481
and
rate limiting https://github.com/microsoft/basis-planning/issues/482

### DIFF
--- a/src/cs/Ssh.Tcp/SshClient.cs
+++ b/src/cs/Ssh.Tcp/SshClient.cs
@@ -105,13 +105,8 @@ public class SshClient : IDisposable
 		tcpClient.Client.ConfigureSocketOptionsForSsh();
 
 		NetworkStream stream = tcpClient.GetStream();
-		var ipEndpoint = tcpClient.Client.RemoteEndPoint as IPEndPoint;
-		IPAddress? ipAddress = null;
-		if (ipEndpoint != null)
-		{
-			ipAddress = ipEndpoint.Address;
-		}
-
+		IPAddress? ipAddress = 
+			tcpClient.Client.RemoteEndPoint is IPEndPoint ipEndpoint? ipEndpoint.Address : null;
 		return (stream, ipAddress);
 	}
 
--- a/src/cs/Ssh.Tcp/SshClient.cs
+++ b/src/cs/Ssh.Tcp/SshClient.cs
@@ -47,7 +47,7 @@ public class SshClient : IDisposable
 		int serverPort = DefaultServerPort,
 		CancellationToken cancellation = default)
 	{
-		var stream = await this.OpenConnectionAsync(
+		(var stream, var ipAddress) = await this.OpenConnectionAsync(
 			serverHost, serverPort, cancellation).ConfigureAwait(false);
 		var session = new SshClientSession(this.config, this.trace);
 
@@ -56,6 +56,7 @@ public class SshClient : IDisposable
 			this.sessions.Add(session);
 		}
 
+		session.RemoteIPAddress = ipAddress;
 		session.Closed += (s, e) =>
 		{
 			if (e.Exception != null)
@@ -87,7 +88,7 @@ public class SshClient : IDisposable
 		return session;
 	}
 
-	protected virtual async Task<Stream> OpenConnectionAsync(
+	protected virtual async Task<(Stream Stream, IPAddress? RemomoteIPAddress)> OpenConnectionAsync(
 		string host, int port, CancellationToken cancellation)
 	{
 #pragma warning disable CA2000 // Dispose objects before losing scope
@@ -104,7 +105,14 @@ public class SshClient : IDisposable
 		tcpClient.Client.ConfigureSocketOptionsForSsh();
 
 		NetworkStream stream = tcpClient.GetStream();
-		return stream;
+		var ipEndpoint = tcpClient.Client.RemoteEndPoint as IPEndPoint;
+		IPAddress? ipAddress = null;
+		if (ipEndpoint != null)
+		{
+			ipAddress = ipEndpoint.Address;
+		}
+
+		return (stream, ipAddress);
 	}
 
 	public async Task ReconnectSessionAsync(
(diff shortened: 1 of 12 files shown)


src/cs/Ssh.Tcp/SshClient.cs
```diff
@@ -104,15 +105,22 @@ await tcpClient.ConnectAsync(host, port)
 		tcpClient.Client.ConfigureSocketOptionsForSsh();
 
 		NetworkStream stream = tcpClient.GetStream();
-		return stream;
+		var ipEndpoint = tcpClient.Client.RemoteEndPoint as IPEndPoint;
```
IlyaBiryukov, line 108: nit: can simplify to
``` csharp
var ipAddress = tcpClient.Client.RemoteEndPoint is IPEndPoint ipEndpoint ? ipEndpoint.Address : null;
```

src/cs/Ssh.Tcp/SshServer.cs
```diff
@@ -198,18 +201,25 @@ await session.CloseAsync(SshDisconnectReason.ProtocolError, ex)
...
 			// The server was disposed.
-			return null;
+			return (null, null);
 		}
 
 		tcpClient.Client.ConfigureSocketOptionsForSsh();
 
 		NetworkStream stream = tcpClient.GetStream();
-		return stream;
+		var ipEndpoint = tcpClient.Client.RemoteEndPoint as IPEndPoint;
```
IlyaBiryukov, line 215: Same here

src/cs/Ssh/IO/SshProtocol.cs
```diff
@@ -160,19 +163,25 @@ internal async Task<string> ReadProtocolVersionAsync(CancellationToken cancellat
 				if (line.StartsWith("SSH-", StringComparison.Ordinal))
 				{
 					this.metrics.AddMessageReceived(i + 1);
-					return line;
+					return (line, proxyIPAddress);
+				}
+				else if (line.StartsWith("PROXY ", StringComparison.Ordinal))
```
IlyaBiryukov, line 168: Are we going to authenticate the other party independently before we trust the proxy version? `ReadProtocolVersion` happens before the encryption and authentication, anyone can snip it if the SSH session runs in a non-protected connection, and a malicious client can fake it. 

In context of dev tunnels, snipping from unprotected channels is less of a concern because of https. However, we must not use this "PROXY" IP protocol version extension for authentication purposes.

src/cs/Ssh/IO/SshProtocol.cs
```diff
@@ -160,19 +163,25 @@ internal async Task<string> ReadProtocolVersionAsync(CancellationToken cancellat
...
+				{
+					var split = line.Split(' ');
```
jasongin, line 170: Would it be useful to also get the destination port from the PROXY protocol here (`split[5]`) and expose that on the session? That would allow for more flexibility in how public port SSH numbers are mapped.

src/cs/Ssh/IO/SshProtocol.cs
jasongin, line 168: Support for PROXY protocol should be conditional on a session configuration setting that is off by default. Otherwise if the server is not behind a proxy then a malicious client could abuse this protocol to lie about their client IP address.

src/cs/Ssh/IO/SshProtocol.cs
```diff
@@ -160,19 +163,25 @@ internal async Task<string> ReadProtocolVersionAsync(CancellationToken cancellat
...
+					if (split.Length > 2 && IPAddress.TryParse(split[2], out var ipAddress))
```
jasongin, line 171: If parsing the PROXY line fails, it will be best to abort the connection (throw a `SshConnectionException`) for security.

jasongin (changes requested):

src/cs/Ssh/IO/SshProtocol.cs
jasongin: Similarly, this must only allow a _single_ PROXY line, to prevent a malicious client from providing their own PROXY line that overwrites the one from the actual proxy.

src/cs/Ssh/IO/SshProtocol.cs
klvnraju: I have done this change in tunnels repository. And will add the destination port there.
Abandoning this PR as we will not be doing this here.

klvnraju:
Closing this PR as we will not be needing it as we are doing it in the service

src/cs/Ssh/IO/SshProtocol.cs
klvnraju: I have done that in tunnels repository. (reading to only single line)

jasongin, line 168: Add a comment linking to the proxy protocol doc: https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt, unless you know of a more authoritative source.

klvnraju: The way I am planning to block users from providing their own ip addresses to override is by not accepting proxy_protocol at nginx level.

jasongin: Do we need to support V2 of the PROXY protocol?

jasongin: If the proxy protocol support is enabled in config and the incoming connection does NOT include a PROXY header, should the connection be aborted?

jasongin: > The way I am planning to block users from providing their own ip addresses to override is by not accepting proxy_protocol at nginx level.

But will nginx pass it through then?

klvnraju: I am aborting the connection in our service if proxy header is not set.

klvnraju: > But will nginx pass it through then?
nginx will not understand or read proxy header it is sent. So, technically there will be two lines in stream (first one our nginx and 2nd one user provided/from the user who is trying to send fake address). We read the first one from our nginx and 2nd one is skipped by our ssh library

PR #8, opened 2022-10-21
Add optional stream factory for MultiChannelStream

jasongin:
This will enable the relay service to use stream subclasses that do extra things like rate-limiting.

PR #9, opened 2022-10-25
Fix MultiChannelStream channel stream factory

jasongin:
This amends the design from the [previous PR](https://github.com/microsoft/dev-tunnels-ssh/pull/8). The pervious approach of a factory function parameter to the constructor made it difficult to provide a useful factory function because the `MultiChannelStream` instance could not yet be referenced at the time it was being constructed. A `protected virtual CreateStream` method is a more flexible design.

### DIFF
--- a/src/cs/Ssh/MultiChannelStream.cs
+++ b/src/cs/Ssh/MultiChannelStream.cs
@@ -38,7 +38,6 @@ namespace Microsoft.DevTunnels.Ssh;
 public class MultiChannelStream : IDisposable
 {
 	private readonly Stream transportStream;
-	private readonly Func<SshChannel, SshStream> streamFactory;
 	private readonly SshSession session;
 
 	private EventHandler<SshChannelOpeningEventArgs>? channelOpening;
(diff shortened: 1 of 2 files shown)


src/cs/Ssh/MultiChannelStream.cs
```diff
@@ -304,6 +287,15 @@ public async Task ConnectAndRunUntilClosedAsync(CancellationToken cancellation =
 		return channel;
 	}
 
+	/// <summary>
+	/// Creates a stream instance for a channel. May be overridden to create a
+	/// <see cref="SshStream" /> subclass.
+	/// </summary>
+	protected virtual SshStream CreateStream(SshChannel channel)
```
IlyaBiryukov, line 294: It's used in async methods and can also be async

IlyaBiryukov (approved):

src/cs/Ssh/MultiChannelStream.cs
jasongin: I can't think of any reason it would be useful to make it async.

PR #11, opened 2022-10-25
Add can-change-port extension

jessetrinity:
This is a followup to the ssh version check change that was done when open sourcing the library. This allows us to not rely on the version name for the port switching behavior.
* Add a new `can-change-port` extension to indicate whether the ssh implementation supports changing ports.
* The `isVsSsh` check (which does this same thing) should be retained for older versions of the library which don't support the extension.
* Removes the `isVsSsh` check where it is no longer needed (for versions `2.x`).
* Bump the minor version.

### DIFF
--- a/src/cs/Ssh.Tcp/Services/LocalPortForwarder.cs
+++ b/src/cs/Ssh.Tcp/Services/LocalPortForwarder.cs
@@ -75,10 +75,10 @@ public class LocalPortForwarder : SshService
 		var listenAddress = LocalIPAddress;
 		try
 		{
-			// When the requested port number is not 0 (random) then changing the port number
-			// is not compliant with the SSH protocol. Only allow it if the remote side is
-			// this library, which is known to support it.
+			// Older versions of this library don't support the can-change-port extension,
+			// so the version string is checked instead.
 			bool canChangePort = LocalPort == 0 ||
+				this.Session.ProtocolExtensions.ContainsKey(SshProtocolExtensionNames.CanChangePort) ||
 				this.pfs.Session.RemoteVersion?.IsVsSsh == true;
 			var trace = this.pfs.Session.Trace;
 			this.listener = await this.pfs.TcpListenerFactory.CreateTcpListenerAsync(
--- a/src/cs/Ssh.Tcp/Services/LocalPortForwarder.cs
+++ b/src/cs/Ssh.Tcp/Services/LocalPortForwarder.cs
@@ -78,7 +78,7 @@ public class LocalPortForwarder : SshService
 			// Older versions of this library don't support the can-change-port extension,
 			// so the version string is checked instead.
 			bool canChangePort = LocalPort == 0 ||
-				this.Session.ProtocolExtensions.ContainsKey(SshProtocolExtensionNames.CanChangePort) ||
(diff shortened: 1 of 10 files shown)


src/ts/ssh/sshSessionConfiguration.ts
```diff
@@ -56,6 +65,7 @@ export class SshSessionConfiguration {
 	public constructor(useSecurity: boolean = true) {
 		this.protocolExtensions.push(SshProtocolExtensionNames.serverSignatureAlgorithms);
 		this.protocolExtensions.push(SshProtocolExtensionNames.openChannelRequest);
+		this.protocolExtensions.push(SshProtocolExtensionNames.canChangePort);
```
jasongin, line 68: Also add this in C# `SshSessionConfiguration.cs`.

src/cs/Ssh.Tcp/Services/LocalPortForwarder.cs
```diff
@@ -75,10 +75,10 @@ internal async Task StartForwardingAsync(CancellationToken cancellation)
 		var listenAddress = LocalIPAddress;
 		try
 		{
-			// When the requested port number is not 0 (random) then changing the port number
-			// is not compliant with the SSH protocol. Only allow it if the remote side is
-			// this library, which is known to support it.
+			// Older versions of this library don't support the can-change-port extension,
+			// so the version string is checked instead.
 			bool canChangePort = LocalPort == 0 ||
+				this.Session.ProtocolExtensions.ContainsKey(SshProtocolExtensionNames.CanChangePort) ||
```
jasongin, line 81: `ProtocolExtensions?.ContainsKey(...) == true`

src/cs/Ssh/Services/KeyExchangeService.cs
```diff
@@ -293,9 +293,6 @@ await Session.HandleMessageAsync(new NewKeysMessage(), cancellation)
 						SshTraceEventIds.AlgorithmNegotiation,
 						$"Client's {nameof(ExchangeContext.KeyExchange)} guess was {guessResult}.");
 					this.exchangeContext.DiscardGuessedInit = !negotiatedKexAlgorthmIsPreferred;
-
-					if (negotiatedKexAlgorthmIsPreferred &&
-						Session.RemoteVersion!.IsVsSsh && Session.RemoteVersion!.Version?.Major < 3)
 					{
 						// VS-SSH v2 had a bug in the logic for determining whether the guess was correct.
 						// Use that alternate logic here to preserve compatibility.
```
jasongin, line 298: This entire block should be deleted, not only the `if` condition.

jasongin (changes requested):

jasongin:
Were you planning to remove the `Name == "Microsoft.DevTunnels.Ssh"` check from `IsVsSsh()`?

jasongin (approved):

jessetrinity:
> Were you planning to remove the `Name == "Microsoft.DevTunnels.Ssh"` check from `IsVsSsh()`?

Won't that cause issues with the (few) already published versions using that name?

jasongin:
> Won't that cause issues with the (few) already published versions using that name?

Yes.

PR #15, opened 2022-11-10
Send kex-init message even with no encryption

jasongin:
This makes `MultiChannelStream` (a simplified SSH session with no encryption) more conformant with the standard SSH protocol, by always sending a `SSH_MSG_KEXINIT` message with `none` for all algorithms immediately after the version handshake.

The change is backward-compatible because the previous code would already accept such a message and validate that it allowed `none` for all the algorithms. That validation is also simplified and optimized in this change.

The performance impact of the additional message should be negligible because neither side is waiting for the message or a response to it, so it does cause any additional delay.

### DIFF
--- a/src/cs/Ssh/SshSession.cs
+++ b/src/cs/Ssh/SshSession.cs
@@ -434,6 +434,11 @@ public class SshSession : IDisposable
 				}
 				else
 				{
+					// When there's no key-exchange service configured, send a key-exchange init message
+					// that specifies "none" for all algorithms.
+					await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
+						.ConfigureAwait(false);
+
 					// When encrypting, the key-exchange step will wait on the version-exchange.
 					// When not encrypting, it must be directly awaited.
 					await this.versionExchangeTask!.WaitAsync(connectCts.Token).ConfigureAwait(false);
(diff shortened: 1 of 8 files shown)


src/cs/Ssh/SshSession.cs
```diff
@@ -899,25 +904,15 @@ await protocol.SendMessageAsync(message, cancellationSource.Token)
...
-			else
-			{
-				await CloseAsync(SshDisconnectReason.KeyExchangeFailed).ConfigureAwait(false);
-				return;
-			}
+			await this.kexService.HandleMessageAsync(message, cancellation).ConfigureAwait(false);
+		}
+		else if (!(message is KeyExchangeInitMessage initMessage && initMessage.AllowsNone))
+		{
+			// The other side required some security, but it's not configured here.
```
IlyaBiryukov, line 913: Should we add a trace too?

IlyaBiryukov (dismissed):

src/cs/Ssh/SshSession.cs
jasongin: The outgoing disconnect message will be traced already, however I see that it doesn't include the reason code. I'll add a `ToString()` override for that.

jasongin: Updated.

IlyaBiryukov (approved):

PR #16, opened 2022-11-12
SecureStream class

jasongin:
Adding `SecureStream`, which provides an end-to-end encrypted two-way authenticated data stream over an underlying transport stream, using the SSH protocol but providing a simplified interface limited to a single duplex stream (channel). It is a complement to `MultiChannelStream`, which is a simplified interface to only the channel-multiplexing part of SSH.

This change also includes some improvements to session and channel close/dispose behaviors.

### DIFF
--- a/src/cs/Ssh/SecureStream.cs
+++ b/src/cs/Ssh/SecureStream.cs
@@ -1,0 +1,291 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.DevTunnels.Ssh.Events;
+
+namespace Microsoft.DevTunnels.Ssh;
+
+/// <summary>
+/// Establishes an end-to-end encrypted two-way authenticated data stream over an underlying
+/// transport stream, using the SSH protocol but providing simplified interface that is limited to
+/// a single duplex stream (channel).
+/// </summary>
+/// <remarks>
+/// This class is a complement to <see cref="MultiChannelStream"/>, which provides only the
+/// channel-multiplexing functions of SSH.
+/// 
+/// To establish a secure connection, the two sides first establish an insecure transport stream
+/// over a pipe, socket, or anything else. Then they encrypt and authenticate the connection
+/// before beginning to send and receive data.
+/// </remarks>
+public class SecureStream : Stream
+{
+	private readonly Stream transportStream;
+	private readonly SshSession session;
+	private readonly SshClientCredentials? clientCredentials;
+	private readonly SshServerCredentials? serverCredentials;
+	private SshStream? stream;
+
+	/// <summary>
+	/// Creates a secure stream over an underlying transport stream, using client credentials
+	/// to authenticate.
+	/// </summary>
+	/// <param name="transportStream">Underlying (insecure) transport stream.</param>
+	/// <param name="clientCredentials">Client authentication credentials.</param>
+	/// <param name="trace">Optional trace source for SSH protocol tracing.</param>
+	public SecureStream(
+		Stream transportStream,
+		SshClientCredentials clientCredentials,
+		TraceSource? trace = null)
+	{
+		this.transportStream = transportStream ??
+			throw new ArgumentNullException(nameof(transportStream));
+		this.session = new SshClientSession(
+			SshSessionConfiguration.Default,
+			trace ?? new TraceSource(nameof(MultiChannelStream)));
+
+		this.session.Authenticating += OnSessionAuthenticating;
+		this.session.Closed += OnSessionClosed;
+
+		this.clientCredentials = clientCredentials ??
+			throw new ArgumentNullException(nameof(clientCredentials));
+	}
+
+	/// <summary>
+	/// Creates a secure stream over an underlying transport stream, using server credentials
+	/// to authenticate.
+	/// </summary>
+	/// <param name="transportStream">Underlying (insecure) transport stream.</param>
+	/// <param name="serverCredentials">Server authentication credentials.</param>
+	/// <param name="trace">Optional trace source for SSH protocol tracing.</param>
+	public SecureStream(
+		Stream transportStream,
+		SshServerCredentials serverCredentials,
+		TraceSource? trace = null)
+	{
+		this.transportStream = transportStream ??
+			throw new ArgumentNullException(nameof(transportStream));
+		this.session = new SshServerSession(
+			SshSessionConfiguration.Default,
+			trace ?? new TraceSource(nameof(MultiChannelStream)));
+
+		this.session.Authenticating += OnSessionAuthenticating;
+		this.session.Closed += OnSessionClosed;
+
+		this.serverCredentials = serverCredentials ??
+			throw new ArgumentNullException(nameof(serverCredentials));
+	}
+
+	public event EventHandler<SshAuthenticatingEventArgs>? Authenticating;
+
+	/// <summary>
+	/// Gets a value indicating whether the session is closed.
+	/// </summary>
+	public bool IsClosed =>
+		this.session.IsClosed;
+
+	/// <summary>
+	/// Event that is rised when underlying ssh session is closed.
+	/// </summary>
+	/// <remarks>
+	/// The event is rised before the session stream is closed.
+	/// The stream will be closed after the event handler.
+	/// </remarks>
+	public EventHandler<SshSessionClosedEventArgs>? Closed
+	{
+		get;
+		set;
+	}
+
+	/// <summary>
+	/// Limits the amount of time that ConnectAsync() may wait for the initial
+	/// session handshake (version exchange).
+	/// </summary>
+	public TimeSpan? ConnectTimeout
+	{
+		get => this.session.ConnectTimeout;
+		set => this.session.ConnectTimeout = value;
+	}
+
+	/// <summary>
+	/// Initiates the SSH session over the transport stream by exchanging initial messages with the
+	/// remote peer. Waits for the protocol version exchange, key exchange, authentication, and
+	/// opened channel. Additional message processing is kicked off as a background task chain.
+	/// </summary>
+	/// <exception cref="SshConnectionException">The connection failed due to a protocol
+	/// error.</exception>
+	/// <exception cref="TimeoutException">The ConnectTimeout property is set and the initial
+	/// version exchange could not be completed within the timeout.</exception>
+	public async Task ConnectAsync(CancellationToken cancellation = default)
+	{
+		if (Authenticating == null)
+		{
+			throw new InvalidOperationException(
+				"An Authenticating event handler must be registered before connecting.");
+		}
+
+		if (this.serverCredentials != null)
+		{
+			var serverSession = (SshServerSession)this.session;
+			serverSession.Credentials = this.serverCredentials;
+		}
+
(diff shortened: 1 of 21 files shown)


src/cs/Ssh/Services/AuthenticationService.cs
```diff
@@ -475,8 +475,14 @@ protected override void Dispose(bool disposing)
 	{
 		if (disposing)
 		{
-			disposeCancellationSource.Cancel();
-			disposeCancellationSource.Dispose();
+			try
+			{
+				disposeCancellationSource.Cancel();
+				disposeCancellationSource.Dispose();
```
IlyaBiryukov, line 481: There are no native resources used in `disposeCancellationSource` (e.g. wait handles)? If so, it's OK to not dispose it.

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+	/// </summary>
+	/// <param name="transportStream">Underlying (insecure) transport stream.</param>
+	/// <param name="clientCredentials">Client authentication credentials.</param>
+	/// <param name="trace">Optional trace source for SSH protocol tracing.</param>
+	public SecureStream(
+		Stream transportStream,
+		SshClientCredentials clientCredentials,
+		TraceSource? trace = null)
+	{
+		this.transportStream = transportStream ??
```
IlyaBiryukov, line 45: Let's also check that `transportStream` can read and write.

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+			throw new ArgumentNullException(nameof(transportStream));
+		this.session = new SshClientSession(
+			SshSessionConfiguration.Default,
+			trace ?? new TraceSource(nameof(MultiChannelStream)));
```
IlyaBiryukov, line 49: `nameof(SecureStream)`

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+	public SecureStream(
+		Stream transportStream,
+		SshServerCredentials serverCredentials,
+		TraceSource? trace = null)
+	{
+		this.transportStream = transportStream ??
+			throw new ArgumentNullException(nameof(transportStream));
+		this.session = new SshServerSession(
+			SshSessionConfiguration.Default,
+			trace ?? new TraceSource(nameof(MultiChannelStream)));
```
IlyaBiryukov, line 74: Same here

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+		Authenticating?.Invoke(this, e);
+	}
+
+	private void OnSessionClosed(object? sender, SshSessionClosedEventArgs e)
+	{
+		this.session.Closed -= OnSessionClosed;
+		Closed?.Invoke(this, e);
+	}
+
+	public override bool CanRead => true;
```
IlyaBiryukov, line 233: Should this check if the session is still active?

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+
+	public override bool CanWrite => true;
```
IlyaBiryukov, line 235: Same here.

src/cs/Ssh/SecureStream.cs
IlyaBiryukov: The session has connected and still active, that is.

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+	/// <param name="trace">Optional trace source for SSH protocol tracing.</param>
+	public SecureStream(
+		Stream transportStream,
+		SshServerCredentials serverCredentials,
+		TraceSource? trace = null)
+	{
+		this.transportStream = transportStream ??
+			throw new ArgumentNullException(nameof(transportStream));
+		this.session = new SshServerSession(
+			SshSessionConfiguration.Default,
```
IlyaBiryukov, line 73: Is reconnection going to be a concern? Currently ssh session from the tunnel client to the tunnel host support reconnection. Will this be supported with the new model?

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+		if (Authenticating == null)
+		{
+			throw new InvalidOperationException(
+				"An Authenticating event handler must be registered before connecting.");
+		}
+
+		if (this.serverCredentials != null)
+		{
+			var serverSession = (SshServerSession)this.session;
+			serverSession.Credentials = this.serverCredentials;
```
IlyaBiryukov, line 134: Why do we assign server credentials here and not in ctor?

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+	public override bool CanWrite => true;
+
+	public override void Flush()
+	{
+		this.stream?.Flush();
+	}
+
+	public override int Read(byte[] buffer, int offset, int count)
+	{
+		var stream = this.stream ?? throw new InvalidOperationException("Stream is not connected.");
```
IlyaBiryukov, line 244: This can be encapsulated into a property

src/ts/ssh/secureStream.ts
```diff
@@ -0,0 +1,235 @@
...
+		credentials: SshClientCredentials | SshServerCredentials,
+	) {
+		super({
+			write(
+				this: SecureStream,
+				chunk: Buffer | string | any,
+				encoding: BufferEncoding,
+				cb: (error?: Error | null) => void,
+			) {
+				if (!this.stream) throw new Error('Stream is not connected.');
```
IlyaBiryukov, line 57: This can be encapsulated in a property

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+/// </summary>
+/// <remarks>
+/// This class is a complement to <see cref="MultiChannelStream"/>, which provides only the
+/// channel-multiplexing functions of SSH.
+/// 
+/// To establish a secure connection, the two sides first establish an insecure transport stream
+/// over a pipe, socket, or anything else. Then they encrypt and authenticate the connection
+/// before beginning to send and receive data.
+/// </remarks>
+public class SecureStream : Stream
```
IlyaBiryukov, line 25: Does this class need `channelMaxWindowSize` for parity with ts?

src/ts/ssh/secureStream.ts
```diff
@@ -0,0 +1,235 @@
...
+
+	public async close() {
+		if (!this.disposed) {
+			this.disposed = true;
+
+			await this.session.close(SshDisconnectReason.none, 'SshSession disposed');
+			this.session.dispose();
+			this.unsubscribe();
+
+			await this.transportStream.close();
```
IlyaBiryukov, line 222: What is the difference between `dispose()` and `close()`, should one of the call the other?

src/cs/Ssh/SecureStream.cs
```diff
@@ -0,0 +1,291 @@
...
+		base.Dispose(disposing);
+	}
+
+	/// <summary>
+	/// Close the SSH session with <see cref="SshDisconnectReason.None"/> reason and dispose the underlying transport stream.
+	/// </summary>
+	public async Task CloseAsync()
+	{
+		await this.session.CloseAsync(SshDisconnectReason.None, this.session.GetType().Name + " disposed").ConfigureAwait(false);
+		this.session.Dispose();
```
IlyaBiryukov, line 207: Should this call `Dispose()`?

test/cs/Ssh.Test/SecureStreamTests.cs
```diff
@@ -0,0 +1,243 @@
...
+			stream.Dispose();
+		}
+	}
+
+	[Theory]
+	[InlineData(true, true)]
+	[InlineData(true, false)]
+	[InlineData(false, true)]
+	[InlineData(false, false)]
+	public async Task DisposeRaisesCloseEvent(bool isConnected, bool disposeAsync)
```
IlyaBiryukov, line 212: Combinatorial test?

IlyaBiryukov (approved):

src/cs/Ssh/SecureStream.cs
jasongin: Disposing the session closes the channel, which will trigger the channel-closed handler that disposes this instance on line 173. I'll add a comment.

src/ts/ssh/secureStream.ts
jasongin: The main difference is that a session that is closed (but not disposed) due to a connection error may be reconnected, if reconnection is enabled in configuration.

Currently SecureStream doesn't enable reconnection, so the additional dispose call here is redundant. But we might add reconnection support later.

src/cs/Ssh/SecureStream.cs
jasongin: Oh, that TypeScript `SecureStream.channelMaxWindowSize` property is not used, It was a mistake from copy-pasting code from `MultiChannelStream`. I'm deleting it from `SecureStream`.

src/cs/Ssh/SecureStream.cs
jasongin: It could be done in either place. But this method is setting up both server and client authentication and is handling client credentials (which cannot be set in the constructor), so I thought it is easier to understand if this method handled the server credentials also.

src/cs/Ssh/SecureStream.cs
jasongin: No, because this property indicates whether the current stream _supports_ reading, not whether the stream is currently in a state that _allows_ reading. https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.canread

src/cs/Ssh/SecureStream.cs
jasongin: Same answer as above.

src/cs/Ssh/SecureStream.cs
jasongin: Not necessary in my opinion, and other stream wrapper classes in this library don't do that. Streams that can't read and write are unusual, and it is sufficient to allow calls to their `Read()`/`Write()` methods to throw `NotSupportedException`.

src/cs/Ssh/Services/AuthenticationService.cs
jasongin: It implements `IDisposable`, so the code analysis will warn if it is not disposed.

src/cs/Ssh/SecureStream.cs
jasongin: I'm working on that, but it will be tricky when some channels are unencrypted and intercepted (potentially even modified?) by the relay, while other channels are E2E-encrypted separately. Therefore all the channels will need to be reconnected separately.

I'll probably add reconnect support to SecureStream later. We'll need a way to reconnect unencrypted channels also... the challenge is without the SSH packet sequencing and HMAC it's difficult to guarantee integrity of the data after reconnection. Maybe we'd use SecureStream with no encryption but with HMAC. (Then the relay could observe but not modify the traffic.)

But not for this PR.

src/cs/Ssh/SecureStream.cs
jasongin: Done.

IlyaBiryukov (approved):

PR #18, opened 2022-11-29
Fix port-forward response message sequencing

jasongin:
When a `PortForwardRequestMessage` is received, the recipient may attempt to immediately open a channel on the newly forwarded port. There was a timing bug in which the `ChannelOpenMessage` could be sent _before_ the `PortForwardSuccessMessage`, and that would cause the other side to reject the channel.

I reproduced the issue by modifying a test case, and fixed it by adding a response continuation mechanism for request events. The `PortForwardingService` uses a response continuation to ensure the port-added event is raised after sending the response, therefore any handler of the port-added event can safely open a channel.

The problem only existed in C#. While TS code follows the same pattern, there was no issue due to the way JS events are raised asynchronously.

Fixes: https://github.com/microsoft/basis-planning/issues/605

### DIFF
--- a/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
+++ b/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
@@ -538,21 +538,22 @@ public class PortForwardingService : SshService
 		}
 
 		request.ResponseTask = Task.FromResult(response ?? new SessionRequestFailureMessage());
-		request.ResponseContinuation = () =>
+
+		if (response is PortForwardSuccessMessage portForwardResponse)
 		{
-			// Add to the collection (and raise event) after sending the response,
-			// to ensure event-handlers can immediately open a channel.
-			if (response is PortForwardSuccessMessage portForwardResponse)
+			request.ResponseContinuation = (_) =>
 			{
+				// Add to the collection (and raise event) after sending the response,
+				// to ensure event-handlers can immediately open a channel.
 				var forwardedPort = new ForwardedPort(
 					localPort: (int)portForwardResponse.Port,
 					remotePort: portForwardRequest.Port == 0 ? null : (int)portForwardRequest.Port,
 					isRemote: true);
 				RemoteForwardedPorts.AddPort(forwardedPort);
-			}
 
-			return Task.CompletedTask;
-		};
+				return Task.CompletedTask;
+			};
+		}
 	}
 
 	private async Task<int?> StartForwardingAsync(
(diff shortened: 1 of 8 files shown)


src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -538,17 +538,21 @@ public PortForwardingService(SshSession session)
 		}
 
 		request.ResponseTask = Task.FromResult(response ?? new SessionRequestFailureMessage());
-
-		// Add to the collection (and raise event) after sending the response,
-		// to ensure event-handlers can immediately open a channel.
-		if (response is PortForwardSuccessMessage portForwardResponse)
+		request.ResponseContinuation = () =>
```
IlyaBiryukov, line 541: As the `response` is known already, we can check it before setting `ResponseContinuation`:
``` csharp
if (response is PortForwardSuccessMessage portForwardResponse)
{
    request.ResponseContinuation = () =>
    {
				var forwardedPort = new ForwardedPort(
					localPort: (int)portForwardResponse.Port,
					remotePort: portForwardRequest.Port == 0 ? null : (int)portForwardRequest.Port,
					isRemote: true);
				RemoteForwardedPorts.AddPort(forwardedPort);
    };
}
```

src/cs/Ssh/Events/SshRequestEventArgs.cs
```diff
@@ -63,6 +64,16 @@ public class SshRequestEventArgs<T> where T : SshMessage
...
 
+	/// <summary>
+	/// Gets or sets an action to be invoked AFTER any response message has been sent.
+	/// (The continuation will be invoked even if a response was not requested.)
+	/// </summary>
+	/// <remarks>
+	/// This enables a request handler to ensure additional messages are sequenced after the
+	/// response message.
+	/// </remarks>
+	public Func<Task>? ResponseContinuation { get; set; }
```
IlyaBiryukov, line 75: It could be `Func<SshMessage, Task>` to make it flexible and handle any result of `ResponseTask`.

IlyaBiryukov (approved):

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Updated.

src/cs/Ssh/Events/SshRequestEventArgs.cs
jasongin: Updated. That requires awaiting the `ResponseTask` even if a response was not requested... but that's probably better anyway to not have an unawaited task.

IlyaBiryukov (approved):

PR #23, opened 2023-01-12
Add .NET Framework 4.8 build target, remove .NET Core 2.1

jasongin:
- Remove .NET Standard 2.0 and .NET Core 2.1 build and test targets, because .NET Core 2.1 is no longer supported.
 - Add .NET Framework 4.8 build and test targets, conditional on Windows OS
 - Change various code to be compatible with .NET Framework.
 - Reduce test failures caused by unavailable crypto algorithms (ECDH, AES-GCM). The crypto tests specific to those algorithms will still fail, but other test cases will fallback to always-available algorithms. This reduces confusion about the cause for many test failures, particularly on Mac OS.
 - Update YAML build definitions to target the appropriate .NET versions.

### DIFF
--- a/src/cs/Directory.Build.props
+++ b/src/cs/Directory.Build.props
@@ -3,9 +3,13 @@
 	<Import Project="../../Directory.Build.props" />
 
 	<PropertyGroup>
-		<TargetFrameworks>netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
+		<TargetFrameworks>netstandard2.1;net6.0</TargetFrameworks>
+		<TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
+		<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net48;$(TargetFrameworks)</TargetFrameworks>
+
 		<IsPackable>true</IsPackable>
 		<NoWarn>$(NoWarn);CS1591</NoWarn><!-- Missing XML comment for publicly visible type or member -->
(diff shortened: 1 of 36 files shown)


jasongin:
.NET Core 3.1 went out of support just last month, and .NET 5 [only has four months left](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core). But I don't see a need to remove those just yet.

IlyaBiryukov (approved):

src/cs/Directory.Build.props
```diff
@@ -3,9 +3,13 @@
 	<Import Project="../../Directory.Build.props" />
 
 	<PropertyGroup>
-		<TargetFrameworks>netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
+		<TargetFrameworks>netstandard2.1;net6.0</TargetFrameworks>
+		<TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
```
taiyosogawa, line 7: Does this just replace the value for `TargetFrameworks` on line 6?

jasongin: Oh, that is a copy-paste mistake. Line 6 (with `netstandard2.1`) is the correct one for the library projects, though it doesn't make a big difference since either `netstandard2.1` or `netcoreapp3.1` can run on both .NET Core 3.1 and .NET 5.

I'll make a PR to fix it.

jasongin: https://github.com/microsoft/dev-tunnels-ssh/pull/27

PR #25, opened 2023-01-18
Add build step to link libssl on Mac OS

jasongin:
This fixes the ECDH and AES-GCM tests that have been failing on Mac OS build workflows because `libssl` was not found.

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -1,4 +1,4 @@
-name: PR Build
+name: SSH Build & Test
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-build.yaml
```diff
@@ -26,6 +26,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      # On the Mac build machines, libssl 1.1 is installed
+      # but not linked in the path expected by .NET. Fix it now.
+      - name: Link libssl on Mac OS
+        run: ln -sn /usr/local/opt/openssl/lib/libssl.1.1.dylib /usr/local/lib/libssl.1.1.dylib
+        if: matrix.os == 'macOS-latest'
```
IlyaBiryukov, line 33: Is this change permanent, does it affect the build agents after the build is over?

IlyaBiryukov, line 33: Can there be a condition that checks if `libssl` is installed?

IlyaBiryukov (approved):

.github/workflows/pr-build.yaml
jasongin: `ln` will fail with nonzero exit code if a file exists at that path; it won't overwrite without the `-f` option.

If that happens the build will fail, and we can address it then. But I don't see a need to make a more complex script now.

jasongin: I'm pretty sure the build agent VMs are reset before every build. If not then the `ln` would fail in subsequent builds (see my other comment).

PR #26, opened 2023-01-18
Fixed server example in docs

anvar-ramazanov:


src/cs/Ssh/README.md
```diff
@@ -117,8 +117,13 @@ server.SessionAuthenticating += (_, e) =>
 };
 
 // Handle channel command requests.
-server.ChannelRequest += (_, e) =>
+server.ChannelRequest += (c, e) =>
 {
+    if (c is not SshChannel channel)
```
jasongin, line 122: Are you seeing `ChannelRequest` events in which the event source parameter is not a `SshChannel`? I didn't think that should be possible, but maybe I'm forgetting something.

anvar-ramazanov: Nope, but in the example `c` has type `object?`. Probably it because of anonymous event handler.

jasongin: It is type `object?` because it's using the [`System.EventHandler<T>` delegate](https://learn.microsoft.com/en-us/dotnet/api/system.eventhandler-1). That's normal and should be familiar to any C# developers. Since the source parameter is not used (`_`), it really doesn't matter and there's no need to check it.

anvar-ramazanov: The problem is that we have a variable `channel` in this event handler, which not exists.

jasongin: Oh I see, thanks for explaining. This change is good then.

Background is that this example used to be structured a bit differently, such that there was a `channel` variable. But I missed that when updating it. Thanks for fixing.

jasongin (approved):

anvar-ramazanov:
Could you please merge it if it looks good to you? I am not allowed to do it

PR #28, opened 2023-01-25
Publish windows .pdb files to symweb and portable .pdb in .snupkg to nuget.org

taiyosogawa:
In order to ensure symbols are available to debug for VS, we need to publish our symbols to symweb in addition to nuget.org.

Symweb [only accepts windows .pdb files](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/672/Archive-Symbols-with-Symweb?anchor=portable-pdbs), so I've added a step to install the [pdb2pdb.exe](https://github.com/dotnet/symreader-converter). I then added a build target that accepts the .exe path as a property and invokes it to convert the newly generated portable .pdb files (which we still need to publish to nuget.org). This new build target generates a file structure like this:

image

Finally, I've added a build step to publish the windows .pdbs to symweb. Doing so requires [setting up a PAT](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/azure-artifacts/symbol-service#cross-account-using-pat).

I verified that this works by running the pipeline with expiration time = 1d. After the pdbs were published, I was able to retrieve them by downloading the pipeline's published artifacts, then running this command:
```
.\SymChk.exe /v /s SRV*https://symweb <path_to_dll_from_downloaded_artifiacts>
```


.pipelines/dev-tunnels-ssh-ci.yaml
```diff
@@ -168,7 +185,7 @@ jobs:
         continueOnError: True
         inputs:
           command: push
-          searchPatternPush: out/pkg/*.nupkg
+          searchPatternPush: out/pkg/*.nupkg;out/pkg/*.snupkg
```
taiyosogawa, line 188: Although I'm not sure how to test this without actually publishing the package to nuget, I am reasonably certain this is the correct command based on these docs:

- [publishing a symbol package](https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg#publishing-a-symbol-package)
- [NuGetCommand@2](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-command-v2?view=azure-pipelines)

IlyaBiryukov (approved):

jasongin (approved):

PR #30, opened 2023-01-30
Add step to build pipeline to use Nuget 6.4.0

taiyosogawa:
Specifying an appropriate version of the nuget CLI so that we can publish .snupkg files.

See this analogous PR: https://github.com/microsoft/dev-tunnels/pull/184

### DIFF
--- a/.pipelines/dev-tunnels-ssh-ci.yaml
+++ b/.pipelines/dev-tunnels-ssh-ci.yaml
@@ -75,6 +75,11 @@ jobs:
         inputs:
           verbose: false
       
+      - task: NuGetToolInstaller@1
+        displayName: Use NuGet 6.4.0
+        inputs:
+          versionSpec: '6.4.0'
+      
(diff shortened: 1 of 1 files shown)


.pipelines/dev-tunnels-ssh-ci.yaml
```diff
@@ -75,6 +75,11 @@ jobs:
         inputs:
           verbose: false
       
+      - task: NuGetToolInstaller@1
+        displayName: Use NuGet 6.4.0
+        inputs:
+          versionSpec: '6.4.0'
```
jasongin, line 81: `'6.4.x'` is probably better, that way it can pick up patch updates which may include security fixes.

jasongin (approved):

.pipelines/dev-tunnels-ssh-ci.yaml
taiyosogawa: Good point - thanks!

PR #36, opened 2023-03-02
Fix port tracking for forwarding listener conflict

jasongin:
The previous way of handling a port-forwarding port number conflict was confusing, a little buggy, and relied on a proprietary "can-change-port" extension. If the server requested port A, but the client found port A was in-use, the client could select port A+1=B instead and _tell the server_ that it is using that port. The server would then have to keep track of the mapping between requested port and actual port numbers. (And not all of that tracking was logically consistent.)

A better way to handle that is when the client chooses a different port number it can just _not tell the server_ about that choice. The client and server continue referring using the requested port number in all messages, and the client takes care of mapping to the actual chosen port number as needed. This simplifies port tracking on the server side and does not require a protocol extension so it's more compatible with other SSH implementations.

The "can-change-port" extension is dropped. That has the following back-compat implications:
 - A new client and an old server (host) will have no issues because the new client can choose a different port without telling the server.
 - An old client used with a new server (host) could fail to forward ports if the requested port number is in-use, because it detects that the server doesn't support that extension.

This scenario only affects the dev tunnels CLI which is not released publicly, so it should be OK to let those older clients break.

### DIFF
--- a/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
+++ b/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
@@ -30,12 +30,38 @@ public class PortForwardingService : SshService
 	internal const string ReversePortForwardChannelType = "direct-tcpip";
 
 	private bool disposed;
+
+	/// <summary>
+	/// Maps from local IP address and FORWARDED port number to the object that manages
+	/// listening for incoming connections on that port and forwarding them through the session.
+	/// </summary>
+	/// <remarks>
+	/// Note the actual local source port number used may be different from the forwarded port
+	/// number if the local TCP listener factory chose a different port. The forwarded port number
+	/// is used to identify the port in any messages exchanged between client and server.
+	/// </remarks>
 	private readonly IDictionary<IPEndPoint, LocalPortForwarder> localForwarders =
 		new Dictionary<IPEndPoint, LocalPortForwarder>();
-	private readonly IDictionary<IPEndPoint, RemotePortConnector> remoteConnectors =
-		new Dictionary<IPEndPoint, RemotePortConnector>();
-	private readonly ICollection<ChannelForwarder> channelForwarders =
-		new List<ChannelForwarder>();
+
+	/// <summary>
+	/// Maps from FORWARDED port numbers to the object that manages relaying forwarded connections
+	/// from the session to a local port.
+	/// </summary>
+	/// <remarks>
+	/// Note the actual local destination port number used may be different from the forwarded port
+	/// number. The forwarded port number is used to identify the port in any messages exchanged
+	/// between client and server.
+	/// </remarks>
+	private readonly IDictionary<int, RemotePortConnector> remoteConnectors =
+		new Dictionary<int, RemotePortConnector>();
+
+	/// <summary>
+	/// Collection of objects that manage forwarding traffic on individual channels.
+	/// </summary>
+	/// <remarks>
+	/// There may be 0 to multiple channel forawarders per forwarded port.
+	/// </remarks>
+	private readonly ICollection<ChannelForwarder> channelForwarders = new List<ChannelForwarder>();
 
 	public PortForwardingService(SshSession session)
 		: base(session)
@@ -195,7 +221,9 @@ public class PortForwardingService : SshService
 		var remoteEndPoint = new IPEndPoint(remoteIPAddress, remotePort);
 		lock (this.remoteConnectors)
 		{
-			this.remoteConnectors.Add(remoteEndPoint, forwarder);
+			// The remote port is the port sent in the message to the other side,
+			// so the connector is indexed on that port number, rather than the local port.
+			this.remoteConnectors.Add(remotePort, forwarder);
 		}
 
 		var forwardedPort = new ForwardedPort(localPort, remotePort, isRemote: false);
@@ -205,7 +233,7 @@ public class PortForwardingService : SshService
 			LocalForwardedPorts.RemovePort(forwardedPort);
 			lock (this.remoteConnectors)
 			{
-				this.remoteConnectors.Remove(remoteEndPoint);
+				this.remoteConnectors.Remove(remotePort);
 			}
 		};
 
@@ -266,17 +294,19 @@ public class PortForwardingService : SshService
 			remotePort);
 		await forwarder.StartForwardingAsync(cancellation).ConfigureAwait(false);
 
-		var localEndPoint = new IPEndPoint(localIPAddress, localPort);
+		// The remote port is the port sent in the message to the other side,
+		// so the forwarder is indexed on that port number, rather than the local port.
+		var forwardingEndPoint = new IPEndPoint(localIPAddress, remotePort);
 		lock (this.localForwarders)
 		{
-			this.localForwarders.Add(localEndPoint, forwarder);
+			this.localForwarders.Add(forwardingEndPoint, forwarder);
 		}
 
 		forwarder.Disposed += (_, _) =>
 		{
 			lock (this.localForwarders)
 			{
-				this.localForwarders.Remove(localEndPoint);
+				this.localForwarders.Remove(forwardingEndPoint);
 			}
 		};
 		return forwarder;
@@ -319,7 +349,9 @@ public class PortForwardingService : SshService
 		var remoteEndPoint = new IPEndPoint(remoteIPAddress, remotePort);
 		lock (this.remoteConnectors)
 		{
-			this.remoteConnectors.Add(remoteEndPoint, streamer);
+			// The remote port is the port sent in the message to the other side,
+			// so the connector is indexed on that port number. (There is no local port anyway.)
+			this.remoteConnectors.Add(remotePort, streamer);
 		}
 
 		var forwardedPort = new ForwardedPort(localPort: null, remotePort, isRemote: false);
@@ -329,7 +361,7 @@ public class PortForwardingService : SshService
 			LocalForwardedPorts.RemovePort(forwardedPort);
 			lock (this.remoteConnectors)
 			{
-				this.localForwarders.Remove(remoteEndPoint);
+				this.remoteConnectors.Remove(remotePort);
 			}
 		};
 
@@ -485,6 +517,7 @@ public class PortForwardingService : SshService
 		var portForwardRequest = request.Request.ConvertTo<PortForwardRequestMessage>();
 
 		SshMessage? response = null;
+		int? localPort = null;
 		var localIPAddress = IPAddressConversions.FromString(portForwardRequest.AddressToBind);
 		if (localIPAddress == null)
 		{
@@ -496,8 +529,6 @@ public class PortForwardingService : SshService
 		}
 		else
 		{
-			var localEndPoint = new IPEndPoint(localIPAddress, (int)portForwardRequest.Port);
-
 			var args = new SshRequestEventArgs<SessionRequestMessage>(
 				request.RequestType!, portForwardRequest, Session.Principal);
 			await base.OnSessionRequestAsync(args, cancellation).ConfigureAwait(false);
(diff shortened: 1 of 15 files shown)


src/ts/ssh-tcp/services/localPortForwarder.ts
```diff
@@ -105,7 +97,6 @@ export class LocalPortForwarder extends SshService {
 					this.tcpListener2 = await this.pfs.tcpListenerFactory.createTcpListener(
 						listenAddress,
 						this.port,
-						false,
```
IlyaBiryukov, line 108: Please also fix the comment above on line 94: `...and not allowing a port change`

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -61,8 +61,25 @@ export class PortForwardingService extends SshService {
...
+
+	/**
+	 * Maps from FORWARDED port numbers to the object that manages relaying forwarded connections
+	 * from the session to a local port.
+	 *
+	 * Note the actual local destination port number used may be different from the forwarded port
+	 * number. The forwarded port number is used to identify the port in any messages exchanged
+	 * between client and server.
+	 */
+	private readonly remoteConnectors = new Map<number, RemotePortConnector>();
```
IlyaBiryukov, line 82: Is `PortForwardingService.cs` already fixed? I don't see it in the PR.

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -239,14 +256,16 @@ export class PortForwardingService extends SshService {
...
+		// The remote port is the port sent in the message to the other side,
+		// so the connector is indexed on that port number, rather than the local port.
+		this.remoteConnectors.set(remotePort, forwarder);
 
 		const forwardedPort = new ForwardedPort(localPort, remotePort, false);
 		this.localForwardedPorts.addPort(forwardedPort);
 		forwarder.onDisposed(() => {
 			this.localForwardedPorts.removePort(forwardedPort);
-			this.remoteConnectors.delete(remoteEndPoint);
+			this.remoteConnectors.delete(remotePort);
```
IlyaBiryukov, line 268: Should we check that `remoteConnectors[remotePort]` is still `forwarder`? In case there is a race where new forwarded port uses the same remote port before the old `forwarder` is disposed.

src/ts/ssh-tcp/services/portForwardingService.ts
jasongin: It is in the PR.

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -61,8 +61,25 @@ export class PortForwardingService extends SshService {
...
+	 *
+	 * Note the actual local source port number used may be different from the forwarded port
+	 * number if the local TCP listener factory chose a different port. The forwarded port number
+	 * is used to identify the port in any messages exchanged between client and server.
+	 */
 	private readonly localForwarders = new Map<string, LocalPortForwarder>();
-	private readonly remoteConnectors = new Map<string, RemotePortConnector>();
+
+	/**
+	 * Maps from FORWARDED port numbers to the object that manages relaying forwarded connections
```
IlyaBiryukov, line 75: `localForwarders` map uses both IP address and forwarded port as the key, and `remoteConnectors` only uses forwarded port. Can it be that the same forwarded port number is forwarded from different local IPs by different local forwarders but would only have one `RemotePortConnector` because `remoteConnectors` key is just the port number?

src/ts/ssh-tcp/services/localPortForwarder.ts
jasongin: Actually this call still should not allow a port change, because it is not expected to listen on different ports between IPv4 and IPv6. I think I'll add the `canChangePort` parameter back to that method.

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -370,14 +391,16 @@ export class PortForwardingService extends SshService {
...
+		// The remote port is the port sent in the message to the other side,
+		// so the connector is indexed on that port number. (There is no local port anyway.)
+		this.remoteConnectors.set(remotePort, streamer);
 
 		const forwardedPort = new ForwardedPort(null, remotePort, false);
 		this.localForwardedPorts.addPort(forwardedPort);
 		streamer.onDisposed(() => {
 			this.localForwardedPorts.removePort(forwardedPort);
-			this.remoteConnectors.delete(remoteEndPoint);
+			this.remoteConnectors.delete(remotePort);
```
IlyaBiryukov, line 403: Same here: should we check whether `remotedConnectors[remotePort]` is still `streamer`?

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -530,26 +552,30 @@ export class PortForwardingService extends SshService {
...
 						cancellation,
 					);
 				} catch (e) {
 					// The error is already traced.
-					localForwardedPort = null;
 				}
-				if (localForwardedPort !== null) {
+				if (localPort !== null) {
+					// The chosen local port may be different from the requested port. Use the
+					// requested port in the response, unless the request was for a random port.
```
IlyaBiryukov, line 569: What does it mean, `request for a random port`? Like  SSH server asks the client, hey, let's forward some random port? What's the user scenario?

src/ts/ssh-tcp/services/portForwardingService.ts
IlyaBiryukov: Found it

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -30,12 +30,38 @@ public class PortForwardingService : SshService
...
+	/// <summary>
+	/// Maps from FORWARDED port numbers to the object that manages relaying forwarded connections
+	/// from the session to a local port.
+	/// </summary>
+	/// <remarks>
+	/// Note the actual local destination port number used may be different from the forwarded port
+	/// number. The forwarded port number is used to identify the port in any messages exchanged
+	/// between client and server.
+	/// </remarks>
+	private readonly IDictionary<int, RemotePortConnector> remoteConnectors =
```
IlyaBiryukov, line 55: Does the key need to be `IPEndPoint` as well, if port forwarding can forward from same port of different interfaces?

src/ts/ssh-tcp/services/portForwardingService.ts
jasongin: Yes. The user scenario is they need to open a separate communication channel between client and server and they don't care what the port number will be - though they will find out what the number is when it's chosen. We don't use that for tunnels currently, but it's part of the protocol.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -506,29 +537,35 @@ public PortForwardingService(SshSession session)
...
 					catch (SocketException)
 					{
 						// The exception is already traced.
-						localPort = null;
 					}
 
 					if (localPort != null)
 					{
+						// The chosen local port may be different from the requested port. Use the
+						// requested port in the response, unless the request was for a random port.
```
IlyaBiryukov, line 556: Why the requested port is not used if the request was for a random port?

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -557,33 +594,40 @@ public PortForwardingService(SshSession session)
...
+				remoteHost: null,
+				remotePort == 0 ? null : remotePort);
 
 			await forwarder.StartForwardingAsync(cancellation).ConfigureAwait(false);
+			localPort = forwarder.LocalPort;
 
-			var forwardingEndPoint = new IPEndPoint(localEndPoint.Address, forwarder.LocalPort);
+			// The remote port is the port received in the message from the other side,
+			// so the forwarder is indexed on that port number, rather than the local port,
+			// unless the request was to choose a random port.
```
IlyaBiryukov, line 623: It seems we may be mixing two namespaces here: remote ports and local ports, and as local ports may be different from remote ports, there may be conflicts. For example
1. Request to choose a random port. Local port 22345 was chosen and used as the key
2. Request to forward with remote port 22345. Some free local port was chosen but we still try to use port 22345 because it's the remote port - Conflict.

src/ts/ssh-tcp/services/portForwardingService.ts
jasongin: > Can it be that the same forwarded port number is forwarded from different local IPs by different local forwarders

It's not possible because it would be ambiguous. After the port is forwarded, when the other side gets a connection it will send a PortForwardChannelOpenMessage with just the port number as the destination. There cannot be multiple destinations for channels associated with the same port number. Anyway the other side doesn't have any information or control over the destination; it is up to the side that requested forwarding to route the incoming channel connections.

And now that I think about it, the `localForwarders` map can use only the port number as key also. I'll update that. It was needlessly complicated before due to an incomplete understanding of the protocol spec when I originally wrote the code.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: That is not supported because the protocol does not have a way to distinguish those when opening channels.

src/ts/ssh-tcp/services/portForwardingService.ts
jasongin: There cannot be a race because the SSH request messages are processed sequentially, even if the processing awaits async operations.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: If the request was 0, indicating a random port, then there was no specific port number in the request. So the actual chosen port number is returned in the response and that number is used in further communications. In that case the resulting "remote" port number is the same as the "local" port number.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: That second request should be rejected because that port number is already forwarded. It doesn't matter whether on the first request it was chosen randomly or specifically.

But you're right that there is no check for that when initially handling the request. I'll add a test case and fix the issue.

src/ts/ssh-tcp/services/localPortForwarder.ts
jasongin: Updated.

src/ts/ssh-tcp/services/portForwardingService.ts
jasongin: Updated `localForwarders` to use a simple port key so the 2 maps are consistent.

src/ts/ssh-tcp/services/portForwardingService.ts
jasongin: Not necessary because processing of received messages is serialized.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Added a check and test case.

IlyaBiryukov:
If there is a breaking change in SSH protocol and the customers need a new version of the dev tunnels CLI, how do we check that in CLI and tell customers to upgrade it?

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -506,29 +546,33 @@ public PortForwardingService(SshSession session)
...
 					}
 
 					if (localPort != null)
 					{
+						// The chosen local port may be different from the requested port. Use the
+						// requested port in the response, unless the request was for a random port.
 						response = new PortForwardSuccessMessage
 						{
-							Port = (uint)localPort.Value,
+							Port = portForwardRequest.Port == 0 ? (uint)localPort.Value :
```
IlyaBiryukov, line 568: If `portForwardRequest.Port` is 0, we use the local port where we're forwarding to. What if that local port matches already used other remote port, will that be a conflict?

For example:
1. Host session forwarded from port 5000
2. Client session couldn't start listening on 5000 because it was in use and started listening on another port. 
3. The app blocking port 5000 on the client exits and the port is now available.
4. Host session wants the client to forward on port 0 and client picks 5000.
5. But 5000 is already used by the port forwarded on step 1-2 because it's the remote port.

IlyaBiryukov (dismissed):

jasongin:
> If there is a breaking change in SSH protocol and the customers need a new version of the dev tunnels CLI, how do we check that in CLI and tell customers to upgrade it?

Each side shares its SSH library version in the initial SSH handshake. If necessary we could check that version. But I don't think that's warranted in this case.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Updated to detect this edge case.

PR #37, opened 2023-03-07
SSH message factory for port-forwarding

jasongin:
A message factory interface for the SSH port-forwarding service will enable the dev-tunnels SDK to extend port-forwarding messages with additional properties to support more functionality and fine-grained access control.

### DIFF
--- a/src/cs/Ssh.Tcp/IPortForwardMessageFactory.cs
+++ b/src/cs/Ssh.Tcp/IPortForwardMessageFactory.cs
@@ -1,0 +1,35 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading.Tasks;
+using Microsoft.DevTunnels.Ssh.Messages;
+
+namespace Microsoft.DevTunnels.Ssh.Tcp;
+
+/// <summary>
+/// Enables applications to extend port-forwarding by providing custom message subclasses
+/// that may include additional properties.
+/// </summary>
+/// <remarks>
+/// Custom message subclasses must override <see cref="SshMessage.OnRead" /> and
+/// <see cref="SshMessage.OnWrite" /> to handle serialization of any additional properties.
+/// </remarks>
+public interface IPortForwardMessageFactory
+{
+	/// <summary>
+	/// Creates a message for requesting to forward a port.
+	/// </summary>
+	/// <returns>An instance or subclass of <see cref="PortForwardRequestMessage" />.</returns>
+	Task<PortForwardRequestMessage> CreateRequestMessageAsync(int port);
+
+	/// <summary>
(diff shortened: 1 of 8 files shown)


src/cs/Ssh.Tcp/IPortForwardMessageFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+/// Custom message subclasses must override <see cref="SshMessage.OnRead" /> and
+/// <see cref="SshMessage.OnWrite" /> to handle serialization of any additional properties.
+/// </remarks>
+public interface IPortForwardMessageFactory
+{
+	/// <summary>
+	/// Creates a message for requesting to forward a port.
+	/// </summary>
+	/// <returns>An instance or subclass of <see cref="PortForwardRequestMessage" />.</returns>
+	Task<PortForwardRequestMessage> CreateRequestMessageAsync(int port);
```
IlyaBiryukov, line 22: Should it be `remotePort`?

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -587,9 +602,12 @@ export class PortForwardingService extends SshService {
 				if (localPort !== null) {
 					// The chosen local port may be different from the requested port. Use the
 					// requested port in the response, unless the request was for a random port.
-					const portResponse = new PortForwardSuccessMessage();
-					portResponse.port =
+					const forwardedPort =
 						portForwardRequest.port === 0 ? localPort : portForwardRequest.port;
+					const portResponse = await this.messageFactory.createSuccessMessageAsync(
```
IlyaBiryukov, line 607: For some reason I cannot find similar place in `PortForwardingService.cs` where `CreateSuccessMessageAsync()` is used.

jasongin: Oh good catch, it was supposed to be called at `PortForwardingService.cs:588`.

(I'm not currently using custom success messages in dev tunnels, but I thought supporting that in the factory could be useful later.)

jasongin: Fixed.

src/cs/Ssh.Tcp/IPortForwardMessageFactory.cs
jasongin: I'm not sure that name will be clear enough in this context. But I'll add doc-comments explaining exactly what that parameter refers to.

IlyaBiryukov (approved):

src/cs/Ssh.Tcp/IPortForwardMessageFactory.cs
jasongin: Updated.

PR #38, opened 2023-03-09
Allow ConnectToForwardedPort to return a forwarded stream without opening a TCP connection

jramsay:
Allow ConnectToForwardedPort to return a forwarded stream without opening TCP connections to a TCP socket. 

- onChannelOpening do not forward channel if acceptLocalConnectionsForForwardedPorts is set to false
- add ConnectToForwardedPortWithoutStartingLocalServer tests

### DIFF
--- a/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
+++ b/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
@@ -788,7 +788,8 @@ public class PortForwardingService : SshService
 
 		request.FailureReason = portForwardRequest.FailureReason;
 		request.FailureDescription = portForwardRequest.FailureDescription;
-		if (request.FailureReason != SshChannelOpenFailureReason.None || !request.IsRemoteRequest)
+		if (request.FailureReason != SshChannelOpenFailureReason.None ||
+			!request.IsRemoteRequest || !AcceptLocalConnectionsForForwardedPorts)
 		{
 			return;
 		}
--- a/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
+++ b/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
@@ -85,6 +85,20 @@ public class PortForwardingService : SshService
 	public bool AcceptLocalConnectionsForForwardedPorts { get; set; } = true;
 
 	/// <summary>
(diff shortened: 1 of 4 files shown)


src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -788,7 +788,8 @@ public PortForwardingService(SshSession session)
 
 		request.FailureReason = portForwardRequest.FailureReason;
 		request.FailureDescription = portForwardRequest.FailureDescription;
-		if (request.FailureReason != SshChannelOpenFailureReason.None || !request.IsRemoteRequest)
+		if (request.FailureReason != SshChannelOpenFailureReason.None ||
+			!request.IsRemoteRequest || !AcceptLocalConnectionsForForwardedPorts)
```
jasongin, line 792: The name of the property `AcceptLocalConnectionsForForwardedPorts` doesn't quite work for this side, where normally the connection would be _forwarded to a local port_, rather than _accepted by listening on a local port_. So it might be better to create a separate property.

jramsay: Thanks, will create a new property: ForwardConnectionsToLocalPort

test/cs/Ssh.Test/PortForwardingTests.cs
```diff
@@ -990,6 +990,30 @@ public async Task ConnectToForwardedPort()
...
+		var clientPfs = this.sessionPair.ClientSession.ActivateService<PortForwardingService>();
+		clientPfs.AcceptLocalConnectionsForForwardedPorts = false;
+
+		var waitTask = this.sessionPair.ServerSession.WaitForForwardedPortAsync(TestPort1)
+			.WithTimeout(Timeout);
+		var forwardTask = this.sessionPair.ClientSession.ForwardFromRemotePortAsync(
+			IPAddress.Loopback, TestPort1).WithTimeout(Timeout);
+		await waitTask;
+
+		var remoteStream = await this.sessionPair.ServerSession
```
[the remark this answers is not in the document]
jramsay: Switched to asserting the local & remote streams and retrieving the local steam from the ChannelOpening

jasongin (approved):

PR #42, opened 2023-03-16
Make channel open response messages extensible

jasongin:
With this change, handlers of the `SshSession.ChannelOpening` event can (asynchronously) return an instance of a subclass of `ChannelOpenConfirmationMessage` or `ChannelOpenFailureMessage` to supply additional information with the response.

This kind of extensibility was already supported for session request and response messages, and channel request messages, so this fills in the gap for channel response messages.

### DIFF
--- a/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
+++ b/src/cs/Ssh.Tcp/Services/PortForwardingService.cs
@@ -726,7 +726,7 @@ public class PortForwardingService : SshService
 		return Task.FromResult(true);
 	}
 
-	protected override async Task OnChannelOpeningAsync(
+	protected override async Task<ChannelMessage> OnChannelOpeningAsync(
 		SshChannelOpeningEventArgs request,
 		CancellationToken cancellation)
 	{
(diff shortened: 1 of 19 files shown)


IlyaBiryukov:
Should this be a major version bump because of the breaking changes?

src/cs/Ssh/Services/SshService.cs
```diff
@@ -101,7 +101,7 @@ protected async Task SendMessageAsync(SshMessage message, CancellationToken canc
 	/// an implementation of this method may add any event-handlers to the
```
IlyaBiryukov, line 101: Please update the xml doc comments on the changed APIs.

IlyaBiryukov (approved):

src/cs/Ssh/Services/SshService.cs
jasongin: I updated comments in the event-args class.

The change in this file is reverted now though. The new revision is simpler overall and more consistent with how other response messages are handled.

IlyaBiryukov (approved):

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -790,46 +790,45 @@ public PortForwardingService(SshSession session)
...
+				var forwardedPort = new ForwardedPort(
+					localPort: remoteForwarder?.LocalPort,
+					remotePort: remoteForwarder?.RemotePort ?? (int)portForwardMessage.Port,
+					isRemote: false);
+				LocalForwardedPorts.AddChannel(forwardedPort, request.Channel);
+			}
+			else
+			{
+				// The forwarding was initiated by the remote session.
+				await RemotePortForwarder.ForwardChannelAsync(
```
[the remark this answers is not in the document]
jasongin: `ForwardConnectionsToLocalPorts` is still checked on line 807. The logic was inverted but the effective behavior should be the same.

PR #45, opened 2023-04-01
Enable stream transforms for port-forwarding

jasongin:
- Add `ForwardedPortConnecting` event on `PortForwardingService`. It enables applications to intercept connections (whether to local sockets or application streams) and optionally apply a transform stream.
 - Rename/refactor `ChannelForwarder` to `StreamForwarder`, using `Stream` APIs to read/write data instead of `SshChannel` messages. This makes it compatible with transform streams.
 - Save the channel request and confirmation messages as properties on the `SshChannel`, so they can be referenced by `ForwardedPortConnecting` event-handlers. (The messages may be subclassed to add application-specific properties.)
 - Improve invalid channel ID tracing in `ConnectionService`.
 - Improve `SecureStream` so that initial attempts to read or write will asynchronously block until the stream is encrypted. (`SecureStream` will be used by the tunnel SDK as a transform stream for port connections.)

This is C# only for now. Porting to TypeScript is tracked by a separate task.


src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -150,6 +151,32 @@ public PortForwardingService(SshSession session)
...
+	public event EventHandler<ForwardedPortConnectingEventArgs>? ForwardedPortConnecting;
+
+	internal async Task<Stream?> OnForwardedPortConnectingAsync(
+		int port,
+		bool isIncoming,
+		SshStream stream,
+		CancellationToken cancellation)
+	{
+		var e = new ForwardedPortConnectingEventArgs(port, isIncoming, stream, cancellation);
+		ForwardedPortConnecting?.Invoke(this, e);
```
IlyaBiryukov, line 167: For most cases the event handler will be `null`, and to save on GC, we can check the handler first:
``` csharp
var handler = ForwardedPortConnecting;
if (handler == null)
{
    return stream;
}

var e = new ForwardedPortConnectingEventArgs(port, isIncoming, stream, cancellation);
handler(this, e);
...
```

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -468,7 +495,16 @@ public PortForwardingService(SshSession session)
 			forwardedPort,
 			cancellation).ConfigureAwait(false);
 
-		return new SshStream(channel);
+		var forwardedStream = await OnForwardedPortConnectingAsync(
+			forwardedPort, isIncoming: false, new SshStream(channel), cancellation)
+			.ConfigureAwait(false);
+		if (forwardedStream == null)
+		{
+			throw new SshChannelException(
```
IlyaBiryukov, line 503: Should we close the SSH channel in this case? Also if the `ForwardedPortConnecting` handler throws.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -893,26 +930,26 @@ await remoteConnector.OnChannelOpeningAsync(request, cancellation)
...
 	{
 		if (this.disposed)
 		{
-			channelForwarder.Dispose();
+			streamForwarder.Dispose();
 		}
 		else
 		{
-			lock (this.channelForwarders)
+			lock (this.streamForwarders)
```
IlyaBiryukov, line 941: The seems to a potential for a race condition here: if `this.disposed` is false when we check for that and then `Dispose()` happens after the `if (this.disposed)` and before `lock (this.streamForwarders)`, the stream forwarder will be added to `this.streamForwarders` and never disposed.

src/cs/Ssh.Tcp/Services/RemotePortForwarder.cs
```diff
@@ -70,17 +72,29 @@ public class RemotePortForwarder : RemotePortConnector
 		// when the PortForwardingService is diposed.
 #pragma warning disable CA2000 // Dispose objects before losing scope
 		var tcpClient = new TcpClient();
+
+		// The event handler may return a transformed channel stream.
+		var forwardedStream = await pfs.OnForwardedPortConnectingAsync(
```
IlyaBiryukov, line 77: Should we catch exceptions in the `ForwardedPortConnecting` handler and reject the request too?

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
```diff
@@ -0,0 +1,214 @@
...
+			Close(abort: !endOfStream);
+		}
+		catch (Exception ex)
+		{
+			// Catch all exceptions in this async void method.
+			Trace.TraceEvent(
+				TraceEventType.Error,
+				SshTraceEventIds.UnknownError,
+				$"{nameof(PortForwardingService)} unexpected error forwarding stream: {ex}");
+			Close(abort: true);
```
IlyaBiryukov, line 51: If `Closed` event handler throws, it will kill the process. It's not caught and inside `async void`.

IlyaBiryukov: Same for `Trace.TraceEvent()`. If trace throws (it can throw `ObjectDisposedException` for example), it'll also kill the process. Async voids are dangerous :)

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
```diff
@@ -0,0 +1,214 @@
...
+			}
+			else
+			{
+				string message = $"Channel forwarder stream read error: {readException.Message}";
+				Trace.TraceEvent(TraceEventType.Verbose, SshTraceEventIds.ChannelClosed, message);
+				return false;
+			}
+		}
+
+		return true;
```
IlyaBiryukov, line 152: Should we flush the destination stream when we're done forwarding?

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
```diff
@@ -0,0 +1,214 @@
...
+	/// Forwards from a source stream to a destination stream until either the end of the source
+	/// stream is reached, an I/O error occurs, or the cancellation token is cancelled.
+	/// </summary>
+	/// <returns>True if the end of stream was reached (graceful closure) else false</returns>
+	private async Task<bool> ForwardStreamAsync(
+		Stream source,
+		Stream destination,
+		CancellationToken cancellation = default)
+	{
+		byte[] buffer = new byte[8192];
```
IlyaBiryukov, line 88: Let's use memory pool [Rent](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.memorypool-1.rent?view=net-7.0) to allocate the buffer.

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
```diff
@@ -0,0 +1,214 @@
...
+		var memory = new Memory<byte>(buffer);
+#endif
+		while (!cancellation.IsCancellationRequested)
+		{
+			int count = 0;
+			Exception? readException = null;
+			try
+			{
+#if SSH_ENABLE_SPAN
+				count = await source.ReadAsync(memory, cancellation)
```
IlyaBiryukov, line 99: We could speed up things a bit if we do reading and writing in parallel after the 1st read:
1. read block 1
2. (in parallel) write block 1 and read block 2
3. repeat (2) until read to the end or exception
4. write the last block and flush

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
```diff
@@ -0,0 +1,214 @@
...
+				string message = $"Channel forwarder stream read error: {readException.Message}";
+				Trace.TraceEvent(TraceEventType.Verbose, SshTraceEventIds.ChannelClosed, message);
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	private void Close(bool abort, string? errorMessage = null)
```
IlyaBiryukov, line 155: The `errorMessage` is never used.

src/cs/Ssh/SecureStream.cs
```diff
@@ -128,50 +129,60 @@ public async Task ConnectAsync(CancellationToken cancellation = default)
...
+			channel.Closed += (_, _) =>
+			{
+				this.Dispose();
+			};
 		}
-		else
+		catch (Exception ex)
 		{
-			channel = await this.session.AcceptChannelAsync(cancellation).ConfigureAwait(false);
+			this.connectCompletion.TrySetException(ex);
```
IlyaBiryukov, line 181: Should we close the session here too? We do close it if authentication fails above.

src/cs/Ssh/SecureStream.cs
```diff
@@ -29,6 +29,7 @@ public class SecureStream : Stream
 	private readonly SshClientCredentials? clientCredentials;
 	private readonly SshServerCredentials? serverCredentials;
 	private SshStream? stream;
+	private readonly TaskCompletionSource<bool> connectCompletion = new ();
```
IlyaBiryukov, line 32: Is the `bool` result of `connectCompletion` used? Can it be just `TaskCompletionSource` ?

src/cs/Ssh/SshChannel.cs
```diff
@@ -115,6 +119,16 @@ public class SshChannel : IDisposable
 
 	public bool IsClosed => this.localClosed || this.remoteClosed;
 
+	/// <summary>
+	/// Gets the message that requested opening the channel.
+	/// </summary>
+	public ChannelOpenMessage OpenMessage { get; private set; }
```
IlyaBiryukov, line 125: `set` is not needed for init-only properties.

src/cs/Ssh/SshStream.cs
```diff
@@ -191,6 +191,8 @@ protected override void Dispose(bool disposing)
 			// Asynchronously close the channel, but don't wait for it.
 			_ = Channel.CloseAsync();
 
+			this.readSemaphore.TryRelease();
```
IlyaBiryukov, line 194: Should we release it after setting `IsDisposed = true;` ? This way the waiting code can check `IsDisposed` reliably after waiting for `readSemaphore`.

IlyaBiryukov (approved):

src/cs/Ssh/SecureStream.cs
jasongin: The non-generic `TaskCompletionSource` is only available starting with .NET 5. But this project still maintains compatibility with .NET Framework.

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
jasongin: - That would require 2 buffers, and would be more complicated code. I hesitate to prematurely optimize.
 - We don't do that anywhere else we're forwarding bytes.
 - The incoming stream is buffered internally already, so I doubt that starting the read into another buffer a little sooner will make a significant difference.
 - Normally the remote side of the connection will be much slower than the local side anyway, so forwarding speed will be limited in that way.

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
jasongin: It is passed to `Channel.CloseAsync()` on line 170. The other side will receive the error message and log it (if the error is not caused by a connection loss).

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Updated.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Fixed, for all the disposable collections in this class.

src/cs/Ssh.Tcp/Services/RemotePortForwarder.cs
jasongin: Updated.

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
jasongin: Updated.

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
jasongin: Done.

src/cs/Ssh.Tcp/Services/StreamForwarder.cs
jasongin: Done.

src/cs/Ssh/SecureStream.cs
jasongin: Updated.

src/cs/Ssh/SshChannel.cs
jasongin: Updated.

src/cs/Ssh/SshStream.cs
jasongin: Updated.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Updated.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -893,53 +955,57 @@ await remoteConnector.OnChannelOpeningAsync(request, cancellation)
...
-			}
+				if (this.disposed) return;
+				this.disposed = true;
 
-			lock (this.localForwarders)
-			{
-				disposables.AddRange(this.localForwarders.Values);
-				this.localForwarders.Clear();
-			}
+				lock (this.streamForwarders)
```
IlyaBiryukov, line 992: It's already locked on line 987: `lock (DisposablesLock)`

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -893,53 +955,57 @@ await remoteConnector.OnChannelOpeningAsync(request, cancellation)
...
+				{
+					disposables.AddRange(this.streamForwarders);
+					this.streamForwarders.Clear();
+				}
 
-			lock (this.remoteConnectors)
-			{
-				disposables.AddRange(this.remoteConnectors.Values);
-				this.remoteConnectors.Clear();
+				lock (this.localForwarders)
```
IlyaBiryukov, line 998: Are these locks used anywhere else? Isn't `lock (DisposablesLock)` enough?

jasongin: Yes, I forgot to delete these...

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
jasongin: Removed.

IlyaBiryukov (approved):

Issue #52, opened 2023-05-01
Piped session hangs when using PipeExtensions.pipeSession


jeanp413:
Hang happens in `forwardSessionRequest`, also when piping a channel data gets corrupted somehow. 
Can repro this consistently on an app I'm writing:
- native ssh client connects to nodejs app
- nodejs app using dev-tunnels-ssh creates local ssh server
  - on new session creates a session (using websocket) to remote ssh server and pipes SSHServerSession to a SSHClientSession
- For data corrupted issue, native ssh client executes some shell script on remote machine but received output is corrupted

Already fixed locally, will create a PR and try to write unit tests than can reproduce this behavior.

jasongin:
@jeanp413 Thanks for reporting the issue and contributing a fix!

I'm curious, how/where are you using the TypeScript SSH library? Is it an open-source project?

jasongin:
I merged the fix. It should be published to npm as version `3.11.7` shortly.

jeanp413:
> I'm curious, how/where are you using the TypeScript SSH library? Is it an open-source project?

It's just a vscode extension that connects to a remote machine using ssh over websocket

> I merged the fix. It should be published to npm as version 3.11.7 shortly.

Thanks :tada: :tada: :tada:

PR #53, opened 2023-05-02
Fixes piped session hangs when using PipeExtensions.pipeSession

jeanp413:
Fixes https://github.com/microsoft/dev-tunnels-ssh/issues/52

Can't figure out how to repro the corrupted data issue in a unit test yet but explanation of my use case and how I hit this bug is in the linked issue

### DIFF
--- a/src/ts/ssh/pipeExtensions.ts
+++ b/src/ts/ssh/pipeExtensions.ts
@@ -114,9 +114,10 @@ export class PipeExtensions {
 		toSession: SshSession,
 		cancellation?: CancellationToken,
 	): Promise<SshMessage> {
-		// `toSession.requestResponse` always set `wantReply` to `true` internally and awaits for response
-		// but `SessionRequestMessage` has an internal cache when piped so it will send original message with `false`,
-		// use `toSession.request` instead so it returns immeadiately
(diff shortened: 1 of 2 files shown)


src/ts/ssh/pipeExtensions.ts
```diff
@@ -168,8 +177,12 @@ export class PipeExtensions {
 		toChannel: SshChannel,
 		data: Buffer,
 	): Promise<void> {
-		await toChannel.send(data, CancellationToken.None);
-		channel.adjustWindow(data.length);
+		// Seems that somehow data gets corrupted/disposed? so do a copy first thing
```
jasongin, line 180: The data buffer is re-used when receiving the next message. In many cases a channel data handler can process the data synchronously without any need for allocating and copying the buffer.

But below, the `send()` method is async because it may need to wait for the channel window to open up before sending the full buffer. So, making a copy of the buffer at this point is the safe thing to do.

It _might_ be possible to optimize such that the copy can be skipped if `send()` manages to complete immediately (which should be the most common case), but I think that would be complicated.

jasongin (approved):
Thanks for the fix!

I'll merge this after updating the comment.

jasongin (approved):

jeanp413:
Thanks for the quick review and merge :100:

PR #58, opened 2023-07-12
SecureStream reconnect and related diagnostics

jasongin:
- Add reconnection APIs to the `SecureStream` class.
 - Add `PortUpdated` event to PFS forwarded ports collection. This makes it possible to "refresh" forwarded ports status after a reconnection, and reconnect `SecureStream` channels that run over forwarded ports.
 - Improve various diagnostic messages related to reconnection.

### DIFF
--- a/src/cs/Ssh.Tcp/Events/ForwardedPortsCollection.cs
+++ b/src/cs/Ssh.Tcp/Events/ForwardedPortsCollection.cs
@@ -64,6 +64,9 @@ public class ForwardedPortsCollection : IReadOnlyCollection<ForwardedPort>
 	/// <summary>Event raised when a port is added to the collection.</summary>
 	public event EventHandler<ForwardedPortEventArgs>? PortAdded;
 
+	/// <summary>Event raised when a port in the collection is updated.</summary>
+	public event EventHandler<ForwardedPortEventArgs>? PortUpdated;
+
 	/// <summary>Event raised when a port is removed from the collection.</summary>
 	public event EventHandler<ForwardedPortEventArgs>? PortRemoved;
 
@@ -73,14 +76,16 @@ public class ForwardedPortsCollection : IReadOnlyCollection<ForwardedPort>
 	/// <summary>Event raised when a channel is removed from the collection.</summary>
 	public event EventHandler<ForwardedPortChannelEventArgs>? PortChannelRemoved;
 
-	internal void AddPort(ForwardedPort port)
+	internal void AddOrUpdatePort(ForwardedPort port)
 	{
-		if (!this.portChannelMap.TryAdd(port, new ConcurrentDictionary<uint, SshChannel>()))
+		if (this.portChannelMap.TryAdd(port, new ConcurrentDictionary<uint, SshChannel>()))
 		{
-			throw new InvalidOperationException($"Port {port} is already in the collection.");
+			PortAdded?.Invoke(this, new ForwardedPortEventArgs(port));
 		}
-
-		PortAdded?.Invoke(this, new ForwardedPortEventArgs(port));
+		else
+		{
+			PortUpdated?.Invoke(this, new ForwardedPortEventArgs(port));
+		}
 	}
 
 	internal void RemovePort(ForwardedPort port)
--- a/src/cs/Ssh.Tcp/Events/ForwardedPortsCollection.cs
+++ b/src/cs/Ssh.Tcp/Events/ForwardedPortsCollection.cs
@@ -64,7 +64,11 @@ public class ForwardedPortsCollection : IReadOnlyCollection<ForwardedPort>
 	/// <summary>Event raised when a port is added to the collection.</summary>
 	public event EventHandler<ForwardedPortEventArgs>? PortAdded;
 
-	/// <summary>Event raised when a port in the collection is updated.</summary>
+	/// <summary>
+	/// Event raised when a port in the collection is updated. "Updating" a port doesn't
+	/// change anything at the SSH protocol level, but the application may use this event
+	/// as a signal to update or refresh its state for the forwarded port.
+	/// </summary>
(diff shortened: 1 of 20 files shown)


src/cs/Ssh/SecureStream.cs
```diff
@@ -206,6 +234,33 @@ public async Task ConnectAsync(CancellationToken cancellation = default)
...
+	public async Task ReconnectAsync(
+		Stream transportStream,
+		CancellationToken cancellation = default)
+	{
+		if (!(this.Session is SshClientSession clientSession))
+		{
+			throw new InvalidOperationException("Cannot reconnect SecureStream server.");
+		}
+
+		this.TransportStream = transportStream ??
```
IlyaBiryukov, line 259: Does the previous transport stream need to be closed?

IlyaBiryukov: Or should we check if the session was connected and throw if not (no reconnecting of a session that haven't connected yet)?

src/cs/Ssh/SshChannel.cs
```diff
@@ -428,7 +428,7 @@ private async Task SendEofAsync(CancellationToken cancellation)
 					Trace.TraceEvent(
 						TraceEventType.Error,
 						SshTraceEventIds.ChannelRequestFailed,
-						$"Channel request failed with exception ${ex.ToString()}.");
+						$"Channel request failed with exception {ex.ToString()}.");
```
IlyaBiryukov, line 431: Nit: just `{ex}`. It calls `.ToString()` on its own. Also if there is any format provider in `Trace`, it will handle exception object better than a string.

src/ts/ssh/secureStream.ts
```diff
@@ -56,6 +61,7 @@ export class SecureStream extends Duplex implements Disposable {
 				cb: (error?: Error | null) => void,
 			): void {
 				this.connectCompletion.promise.then((stream) => {
+					if (!stream) throw new ObjectDisposedError('SecureStream');
```
IlyaBiryukov, line 64: Should that be 
``` ts
if (!stream) {
    cb(new ObjectDisposedError('SecureStream'));
} else {
    stream._write(chunk, encoding, cb);
}
```
```

File: src/ts/ssh/secureStream.ts [hunk 4]
```diff
@@ -182,12 +217,41 @@ export class SecureStream extends Duplex implements Disposable {
...
+	 * Applies only to a secure stream client. (The secure stream server handles reconnections
+	 * automatically during the session handshake.)
+	 */
+	public async reconnect(transportStream: Stream | Duplex, cancellation?: CancellationToken) {
+		if (!(this.session instanceof SshClientSession)) {
+			throw new Error('Cannot reconnect SecureStream server.');
+		}
+
+		if (!transportStream) throw new TypeError('A transport stream is required.');
+		this.transportStream = transportStream;
```
IlyaBiryukov, line 248: Should we check if the session is disconnected before trying to reconnect? Also should we close the previous `transportStream`?

IlyaBiryukov (approved):

src/cs/Ssh/SecureStream.cs
jasongin: > Does the previous transport stream need to be closed?

In a disconnected state, the previous stream would already be closed.

> should we check if the session was connected and throw if not

`clientSession.ReconnectAsync()` on line 261 does that.

src/ts/ssh/secureStream.ts
jasongin: In a disconnected state, the previous stream would already be closed.

`this.session.reconnect()` will throw if not in disconnected state.

IlyaBiryukov (dismissed):

taiyosogawa (commented):
The demo for this was very neat! Thanks for the helpful comments where appropriate.

src/cs/Ssh.Tcp/Events/ForwardedPortsCollection.cs
```diff
@@ -73,14 +76,16 @@ public IEnumerable<SshChannel> GetChannels(ForwardedPort port)
...
 	{
-		if (!this.portChannelMap.TryAdd(port, new ConcurrentDictionary<uint, SshChannel>()))
+		if (this.portChannelMap.TryAdd(port, new ConcurrentDictionary<uint, SshChannel>()))
 		{
-			throw new InvalidOperationException($"Port {port} is already in the collection.");
+			PortAdded?.Invoke(this, new ForwardedPortEventArgs(port));
+		}
+		else
+		{
+			PortUpdated?.Invoke(this, new ForwardedPortEventArgs(port));
```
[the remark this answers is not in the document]
jasongin: Right, the update doesn't do anything here. It is just a signal to the application to update or refresh the state of the forwarded port. So it's up to the application to decide what that "update" means, if anything. (In the tunnel SDK I use it to reconnect SecureStream channels after the outer session connection has been restored.)

I'll add more comments to the `PortUpdated` event above.

src/cs/Ssh.Tcp/Events/ForwardedPortsCollection.cs
```diff
@@ -73,14 +76,16 @@ public IEnumerable<SshChannel> GetChannels(ForwardedPort port)
 	/// <summary>Event raised when a channel is removed from the collection.</summary>
 	public event EventHandler<ForwardedPortChannelEventArgs>? PortChannelRemoved;
 
-	internal void AddPort(ForwardedPort port)
+	internal void AddOrUpdatePort(ForwardedPort port)
```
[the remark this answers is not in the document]
jasongin: I still think of it as an "update" so I'll keep the name. But I added comments to the `PortUpdated` event so hopefully that clarifies things.

taiyosogawa (approved):

PR #61, opened 2023-07-25
Fix reconnect timing bug

jasongin:
Reconnection could result in dropped or out-of-order packets sometimes, if the SSH client was trying to send packets at the same time as the reconnect protocol negotiation. The fix is to defer any other messages being sent until after the reconnect has completed (the same way messages are deferred during key-exchange).

The problem did not occur if only the server was sending messages during reconnection, due to how reconnection on the server side swaps out the `SshServerSession` instance being reconnected, unlike the client reconnection where the same session instance is used the whole time.

The problem also did not occur in TypeScript due to the single-threaded execution environment there. However, I still ported some of the refactoring from C# to TS to keep the codebases consistent.

### DIFF
--- a/src/ts/ssh/services/keyExchangeService.ts
+++ b/src/ts/ssh/services/keyExchangeService.ts
@@ -49,6 +49,7 @@ class ExchangeContext {
 	public exchangeValue?: Buffer;
 	public exchange?: KeyExchange;
 	public newAlgorithms?: SshSessionAlgorithms;
+	public isExtensionInfoRequested?: boolean;
 }
 
 const serverExtensionInfoSignal = 'ext-info-s';
@@ -94,9 +95,24 @@ export class KeyExchangeService extends SshService {
 	}
 
 	public finishKeyExchange(): SshSessionAlgorithms {
+		if (!this.exchangeContext) {
+			throw new SshConnectionError(
+				'Key exchange not started.',
+				SshDisconnectReason.protocolError,
+			);
+		}
+
 		const newAlgorithms = this.exchangeContext!.newAlgorithms;
+		if (!newAlgorithms) {
+			throw new SshConnectionError(
+				'Key exchange not completed.',
+				SshDisconnectReason.protocolError,
+			);
+		}
+
+		newAlgorithms.isExtensionInfoRequested = this.exchangeContext?.isExtensionInfoRequested;
 		this.exchangeContext = null;
-		return <SshSessionAlgorithms>newAlgorithms;
+		return newAlgorithms;
 	}
 
 	public abortKeyExchange(): void {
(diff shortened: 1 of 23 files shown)


src/ts/ssh/services/keyExchangeService.ts
```diff
@@ -262,46 +276,50 @@ export class KeyExchangeService extends SshService {
...
-			extensionInfoSignal = clientExtensionInfoSignal;
-		}
-
-		if (this.isInitialExchange && message.keyExchangeAlgorithms!.includes(extensionInfoSignal)) {
-			// The extension info message will be blocked in the queue
-			// until immediately after the key-exchange is done.
-			await this.session.sendExtensionInfo(cancellation);
+			this.exchangeContext.isExtensionInfoRequested =
+				this.isInitialExchange &&
+				message.keyExchangeAlgorithms?.includes(clientExtensionInfoSignal);
```
taiyosogawa, line 322: Same as above.

src/ts/ssh/services/keyExchangeService.ts
```diff
@@ -262,46 +276,50 @@ export class KeyExchangeService extends SshService {
...
+					TraceLevel.Verbose,
+					SshTraceEventIds.algorithmNegotiation,
+					'Already sent correct guess for key-exchange init.',
+				);
 			}
 
-			extensionInfoSignal = serverExtensionInfoSignal;
+			this.exchangeContext.isExtensionInfoRequested =
+				this.isInitialExchange &&
+				message.keyExchangeAlgorithms?.includes(serverExtensionInfoSignal);
```
taiyosogawa, line 302: Should we add `== true` here because `message.keyExchangeAlgorithms` could be null? For all intents and purposes, I'm guessing null is treated the same as false, but it might be good to make it explicitly false.

src/ts/ssh/sshServerSession.ts
```diff
@@ -253,14 +253,19 @@ export class SshServerSession extends SshSession {
 		reconnectSession.reconnectedEmitter.fire();
 	}
 
-	public dispose(): void {
+	public dispose(): void;
+
+	/* @internal */
+	public dispose(error?: Error): void;
```
taiyosogawa, line 259: Does this add something distinct from the `dispose(error?: Error)` below?

test/ts/ssh-test/reconnectTests.ts
```diff
@@ -747,4 +755,111 @@ export class ReconnectTests {
...
+								stream.once('drain', resolve);
+							}
+						}),
+						streamErrorPromise,
+					]);
+					sent = true;
+
+					await Promise.race([readPromise, streamErrorPromise]);
+					const reader = new SshDataReader(receiveBuffer);
+					assert.strictEqual(reader.readUInt32(), receiveCount);
```
taiyosogawa, line 807: Do we want to do anything to verify how many times this is called? I'm just thinking about testing scenarios where the assert is skipped because something above it changed the execution flow. I don't want to make this too complicated - just checking if it's important.

src/ts/ssh/sshServerSession.ts
jasongin: This is how method overloads are declared in TypeScript. First all the overloads are listed with no body, then there is one implementation for all of them.

src/ts/ssh/services/keyExchangeService.ts
jasongin: Conventional style in JS/TS is to rely on null/undefined being _falsy_ in boolean expressions, without being explicit about it.

taiyosogawa (approved):

src/ts/ssh/services/keyExchangeService.ts
jasongin: It is very conventional and idiomatic in JS/TS to rely on boolean coercion in expressions.

test/ts/ssh-test/reconnectTests.ts
jasongin: The exact number of times will vary depending on timing. There are assertions below (lines 841-842, 854-855) that verify at least _some_ messages were sent/received both before and after reconnection.

PR #62, opened 2023-07-26
Fix port-forwarding to IPv6 localhost

jasongin:
When forwarding a port, the server listening on that port might be bound to only IPv4 localhost (`127.0.0.1`), only IPv6 localhost (`::1`), or both. The SSH port-forwarding APIs allow the application to specify the host for forwarded connections as a specific IP address (such as `127.0.0.1` or `::1`) or as a hostname (such as `localhost`). If the application doesn't know whether the server is listening on IPv4 or IPv6, then it should be able to specify `localhost` to support either/both.

.NET TCP sockets on Windows do not support that very well: if the server is only listening on IPv4, there is **2-second delay** while the `::1` connection attempt times out before it falls back to successfully connecting to `127.0.0.1`. I consider that 2-second delay to be unacceptable performance for our SSH port-forwarding scenarios.

There is a conventional approach to handling this, defined in RFC 8305, but .NET has not implemented it yet. So this PR implements a very basic version of that algorithm, which works well at least for the IPv4/IPv6 localhost scenario. (It can be applied more generally to any hostnames that resolve to multiple IPv4 and/or IPv6 DNS records, but we're not concerned about that here.)

Node.js implemented a similar solution starting with v18. So I can just enable that option when creating the socket connection there.

I also created benchmarks for different combinations of IPv4/IPv6/localhost clients and hosts, to verify the performance of both .NET and Node.js SSH port-forwarding.

### DIFF
--- a/src/cs/Ssh.Tcp/SshServer.cs
+++ b/src/cs/Ssh.Tcp/SshServer.cs
@@ -222,7 +222,15 @@ public class SshServer : IDisposable
 	{
 		if (disposing)
 		{
-			this.disposeCancellationSource.Cancel();
+			try
+			{
+				this.disposeCancellationSource.Cancel();
+			}
+			catch (ObjectDisposedException)
+			{
+				// The cancellation source was already disposed.
+			}
+			
 			this.disposeCancellationSource.Dispose();
 
 			foreach (SshServerSession serverSession in this.sessions.ToArray())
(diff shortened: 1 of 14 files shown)


src/cs/Ssh.Tcp/SshServer.cs
```diff
@@ -222,7 +222,15 @@ protected virtual void Dispose(bool disposing)
 	{
 		if (disposing)
 		{
-			this.disposeCancellationSource.Cancel();
+			try
+			{
+				this.disposeCancellationSource.Cancel();
```
IlyaBiryukov, line 227: If the code doesn't use wait handle from `disposeCancellationSource`, there is no need to dispose it and catch an exception here.

IlyaBiryukov (approved):

PR #67, opened 2023-07-31
Handle `SSH_MSG_CHANNEL_EXTENDED_DATA` msg type

mustard-mh:
See [RFC](https://www.ietf.org/rfc/rfc4254.txt#:~:text=be%20passed%20with-,SSH_MSG_CHANNEL_EXTENDED_DATA,-messages%2C%20where%20a), search SSH_MSG_CHANNEL_EXTENDED_DATA

Fix https://github.com/microsoft/dev-tunnels-ssh/issues/68

### DIFF
--- a/src/ts/ssh/sshChannel.ts
+++ b/src/ts/ssh/sshChannel.ts
@@ -30,6 +30,7 @@ import { CancellationToken, CancellationError, withCancellation } from './util/c
 import { Semaphore } from './util/semaphore';
 import { TraceLevel, SshTraceEventIds } from './trace';
 import { PipeExtensions } from './pipeExtensions';
+import { SshExtendedDataEventArgs, SshExtendedDataType } from './events/sshExtendedDataEventArgs';
 
 /**
  * Represents a channel on an SSH session. A session may include multiple channels, which
@@ -74,7 +75,7 @@ export class SshChannel implements Disposable {
 
 	private readonly dataReceivedEmitter = new Emitter<Buffer>();
 
-	private readonly extendedDataReceivedEmitter = new Emitter<Buffer>();
+	private readonly extendedDataReceivedEmitter = new Emitter<SshExtendedDataEventArgs>();
 
 	/**
 	 * Event raised when a data message is received on the channel.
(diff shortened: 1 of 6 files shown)


src/ts/ssh/sshChannel.ts
```diff
@@ -271,6 +276,61 @@ export class SshChannel implements Disposable {
 		}
 	}
 
+	public async sendStderr(data: Buffer, cancellation?: CancellationToken): Promise<void> {
```
IlyaBiryukov, line 279: How about making it flexible: name it `sendExtendedData()` and make `dataTypeCode` an argument?

IlyaBiryukov (dismissed):

src/ts/ssh/sshChannel.ts
IlyaBiryukov: `sendStderr` may then just call it with `1` data type code.

mustard-mh: @IlyaBiryukov , I pushed changes (see [diff](https://github.com/microsoft/dev-tunnels-ssh/compare/eb479a142f45dc8c7e6a7ace4da2b687280498ae...bede4e0942645b087614102ad703861485d43120?expand=1)) to address this comment

src/ts/ssh/sshChannel.ts
```diff
@@ -388,6 +449,11 @@ export class SshChannel implements Disposable {
 		this.dataReceivedEmitter.fire(data);
 	}
 
+	public handleExtendedDataReceived(data: SshExtendedDataEventArgs): void {
+		// this.metrics.addBytesReceived(data.length);
```
jasongin, line 453: Extended data should be included in channel metrics.

src/ts/ssh/sshChannel.ts
```diff
@@ -271,6 +277,61 @@ export class SshChannel implements Disposable {
 		}
 	}
 
+	public async sendExtendedData(dataTypeCode: SshExtendedDataType, data: Buffer, cancellation?: CancellationToken): Promise<void> {
```
jasongin, line 280: Would it be possible to refactor to eliminate some of the duplication of code between this method and `send()` above?

jasongin (commented):
Thanks for the PR! It would be great to implement this also in the C# lib. But that can be done separately.

src/ts/ssh/sshChannel.ts
IlyaBiryukov: Thanks! The change looks good for me. I'll defer to Jason to sign off, he has more comments.

src/ts/ssh/sshChannel.ts
mustard-mh: @jasongin  Addressed, but don't know if you like `sendCommon` function or not 🙏

jasongin:
Published to npm as version `3.11.24`.

PR #69, opened 2023-08-01
Adding auto-comment GitHub action that explains the steps for CLI to pick up ssh repo changes

setaskin:
Adding a reminder comment in PRs that shows necessary steps to be taken for CLi to pick up changes.

### DIFF
--- a/.github/workflows/auto-comment.yaml
+++ b/.github/workflows/auto-comment.yaml
@@ -9,14 +9,14 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           pullRequestOpened: |
             👋 @{{ author }}
-            Once you merge this PR, please make sure you have follow these steps for CLI to pick up your changes. 
+            Once you merge this PR, please make sure you have follow these steps for CLI to pick up your changes.
             1. Go to SSH CI build [link]({{ https://dev.azure.com/devdiv/OnlineServices/_build?definitionId=17611&_a=summary }}).
-            2. Rerun the build by targeting "main" branch with "PublishNpmPackage" and "PublishNugetPAckage" check boxes checked. 
-            3. Once build is completed, the build number is your new ssh package number. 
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-comment.yaml
```diff
@@ -0,0 +1,22 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          pullRequestOpened: |
+            👋 @{{ author }}
+            Once you merge this PR, please make sure you have follow these steps for CLI to pick up your changes.
+            1. Go to [SSH CI build]({{ https://dev.azure.com/devdiv/OnlineServices/_build?definitionId=17611&_a=summary }}).
+            2. Rerun the build by targeting "main" branch with "PublishNpmPackage" and "PublishNugetPackage" check boxes checked.
+            3. Once build is completed, the build number is your new ssh package number.
+            4. Go to SDK repo and send a new Pr with following changes made:
+              4.1 Update the "DevTunnelSshPackageVersion" [here in SDK repo]({{ https://github.com/microsoft/dev-tunnels/blob/main/cs/build/build.props }})
+              4.2 Update all "microsoft/dev-tunnels-ssh" and "microsoft/dev-tunnels-ssh-tcp" depenencies in the sdk repo with new version number.
```
IlyaBiryukov, line 18: After 4.2, should we run `npm install`  to update the `ts/package-lock.json` file in the SDK repo?

IlyaBiryukov (approved):

.github/workflows/auto-comment.yaml
setaskin: Added as step 4.3

jasongin (changes requested):
I don't think it is appropriate to refer to internal-only repos and processes in a public repo.

setaskin:
> I don't think it is appropriate to refer to internal-only repos and processes in a public repo.

that is right. Maybe I can just  create an internal developer document and just mentioning the document name here? Something like "please check out how to update package version documentatio "?  without publishing detailed steps here ??  or do you have any suggestions?

PR #76, opened 2023-08-28
Implement SSH interactive auth method

jasongin:
- Implement the `keyboard-interactive` SSH authentication protocol as defined in the RFC
   - Add `AuthenticationInfoRequest`, `AuthenticationInfoResponse`, and related properties to `SshAuthenticatingEventArgs`.
   - SSH server and client applications may handle the `Authenticating` event and use the new properties to participate in interactive authentication. The library itself does not implement any direct interactivity via console or otherwise.
 - Add the ability to configure enabled auth methods for a session. Disabling unused auth methods can make the auth exchange a little more efficient, and reduce attack surface.
 - Add the concept of "contextual" message number to message type mappings in config. This was necessary because authentication message numbers can map to different types depending on the current authentication request.
   - Specifically, `AuthenticationInfoRequest` and `PublicKeyOKMessage` are both `60`.
   - Technically other key-exchange methods may also overlap with DH message numbers, but the library only supports DH now (and ECDH which uses the same message types).

### DIFF
--- a/src/cs/Ssh/Events/SshAuthenticatingEventArgs.cs
+++ b/src/cs/Ssh/Events/SshAuthenticatingEventArgs.cs
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DevTunnels.Ssh.Algorithms;
+using Microsoft.DevTunnels.Ssh.Messages;
 
 namespace Microsoft.DevTunnels.Ssh.Events;
 
@@ -81,6 +83,16 @@ public enum SshAuthenticationType
 	ClientPublicKey = 4,
 
 	/// <summary>
+	/// The client is attempting to authenticate with interactive prompts.
+	/// </summary>
+	/// <remarks>
+	/// This event is raised by an <see cref="SshServerSession"/> when the client requests
+	/// authentication using the "keyboard-interactive" method. The event may be raised multiple
+	/// times for the same client to facilitate multi-step authentication.
+	/// </remarks>
+	ClientInteractive = 5,
+
+	/// <summary>
 	/// The server is attempting to authenticate with a public key credential.
 	/// </summary>
 	/// <remarks>
@@ -106,7 +118,7 @@ public enum SshAuthenticationType
 /// After validating the credentials, the event handler must set the
 /// <see cref="AuthenticationTask" /> property to a task that resolves to a principal object
 /// to indicate successful authentication. That principal will then be associated with the
-/// sesssion as the <see cref="SshSession.Principal" /> property.
+/// session as the <see cref="SshSession.Principal" /> property.
 /// </remarks>
 [DebuggerDisplay("{ToString(),nq}")]
 [DebuggerStepThrough]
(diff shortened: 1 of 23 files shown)


src/cs/Ssh/Events/SshAuthenticatingEventArgs.cs
```diff
@@ -199,6 +231,22 @@ public class SshAuthenticatingEventArgs
 	/// </summary>
 	public string? ClientUsername { get; }
 
+	/// <summary>
+	/// Gets or sets a request more information for interactive authentication.
```
IlyaBiryukov, line 235: a request *for* more information ?

src/cs/Ssh/Services/AuthenticationService.cs
```diff
@@ -171,17 +199,62 @@ await HandleMessageAsync(message.ConvertTo<PasswordRequestMessage>(), cancellati
...
+			username: null,
+			infoRequest: message,
+			infoResponse: null);
+		await HandleAuthenticatingAsync(args, cancellation).ConfigureAwait(false);
+	}
+
+	private async Task HandleMessageAsync(
+		AuthenticationInfoResponseMessage message, CancellationToken cancellation)
+	{
+		// Raise an Authenticating event that allows the server to process the client's responses
```
IlyaBiryukov, line 236: How will this behave if `Authenticating` event handler is not set? Will it just hang never responding back?

IlyaBiryukov: I see that it would fail auth, good.

src/cs/Ssh/Services/AuthenticationService.cs
```diff
@@ -273,89 +363,128 @@ await Session.SendMessageAsync(new AuthenticationSuccessMessage(), cancellation)
...
+				this.clientAuthenticationMethods.Enqueue(
+					(AuthenticationMethods.Password, async (cancellation) =>
+					{
+						await RequestAuthenticationAsync(username, password, cancellation)
+							.ConfigureAwait(false);
+					}));
+			}
 		}
 
+		// Only add None or Interactive methods if no client credentials were supplied.
```
klvnraju, line 448: I did not understand this comment. Do we try password authentication before keyboard interactive?
If the server supports both keyboard interactive and password the standard ssh client will try password authentication before keyboard interactive?

src/ts/ssh/sshSession.ts
```diff
@@ -170,6 +170,8 @@ export class SshSession implements Disposable {
 	public trace: Trace = (level, eventId, msg, err) => {};
 
 	public constructor(public readonly config: SshSessionConfiguration, isClientSession?: boolean) {
+		this.isClientSession = isClientSession ?? false;
```
IlyaBiryukov, line 173: nit: `this.isClientSession = !! isClientSession`

IlyaBiryukov (approved):

src/cs/Ssh/Services/AuthenticationService.cs
jasongin: This code runs on the client side. The client application may supply public-key or password `SshClientCredentials` ahead of time, before connecting/authenticating the session; in that case only the supplied credentials are used and there is no fallback to other methods. Otherwise the client will fallback to `none` and/or `keyboard-interactive` methods, if the server indicates support for those. It currently does not ever fallback to password auth, though maybe it could.

klvnraju (dismissed):

src/cs/Ssh/Events/SshAuthenticatingEventArgs.cs
jasongin: Fixed.

src/ts/ssh/sshSession.ts
jasongin: Updated to make the property allow `undefined`, because it is possible to construct an `SshSession` that is neither specifically a client or a server (when not doing any encryption/auth).

klvnraju (approved):

PR #82, opened 2023-09-12
Fix tests on Node.js v20

jasongin:
- Node.js v18 added web crypto APIs (alongside node crypto APIs), which broke the logic that switched between web and node crypto APIs.
 - Node.js v20 changed behavior of `setImmediate()` in a way that broke one of the tests that was making an assumption about that behavior.
 - `eslint-plugin-jsdoc` package was complaining about v20 engine not supported. Fixed by updating the package version.
 - PR/CI builds are updated to test with Node v20. Manual testing still passed on Node v16, v18, and in browser.


src/ts/ssh-keys/pkcs1KeyFormatter.ts
```diff
@@ -206,7 +206,11 @@ export class Pkcs1KeyFormatter implements KeyFormatter {
 		iv: Buffer,
 		keyLength: number,
 	): Buffer {
-		const useWebCrypto = !!(typeof crypto === 'object' && crypto.subtle);
+		//console.dir(crypto);
```
IlyaBiryukov, line 209: Nit: please remove commented code.

src/ts/ssh-keys/pkcs8KeyFormatter.ts
```diff
@@ -516,7 +516,8 @@ export class Pkcs8KeyFormatter implements KeyFormatter {
 		iterations: number,
 		keyLength: number,
 	): Promise<Buffer> {
-		const useWebCrypto = !!(typeof crypto === 'object' && crypto.subtle);
+		const useWebCrypto = typeof window !== 'undefined' &&
```
IlyaBiryukov, line 519: Refactor into a helper function? It's also used in other place.

IlyaBiryukov (approved):

src/ts/ssh-keys/pkcs8KeyFormatter.ts
jasongin: Done for 2 of them. The 3rd is in another package so they can't share without making it a public API.

IlyaBiryukov (approved):

PR #88, opened 2023-12-13
Add report progress infrastructure

jramsay:


jramsay:
I'll follow with another PR for other languages. Wanted to get this in for Typescript first so that Codespaces can start consuming it. 

Similar to trace you can set a function that reports connection progress events associated with the session. I've tested against Live Share and I'm able to track what is going on in various scenarios: MultiChannelRpcServer connection, MulitChannelRpcClient connection, Host Share, Host accepting guest connection, Guest joining host connection.

jasongin (changes requested):
See my comments about the progress-reporting design.

src/ts/ssh/progress.ts
```diff
@@ -0,0 +1,163 @@
...
+import {
+	ExtensionInfoMessage,
+	ServiceAcceptMessage,
+	ServiceRequestMessage,
+	SessionRequestSuccessMessage } from "./messages/transportMessages";
+
+/**
+ * Connection progress
+ */
+export enum Progress {
```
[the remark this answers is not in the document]
jasongin: The version exchange is one of the most important steps to report on because it validates the 2 sides are communicating at all. It does not use the standard message protocol so it needs to be handled as a special case.

src/ts/ssh/sshSession.ts
```diff
@@ -332,6 +346,7 @@ export class SshSession implements Disposable {
 			// When encrypting, the key-exchange step will wait on the version-exchange.
 			// When not encrypting, it must be directly awaited.
 			await withCancellation(this.versionExchangePromise!, cancellation);
+			this.raiseReportProgress(Progress.ReceivingProtocolVersionExchange);
```
jasongin, line 349: Should it be "Completed" instead of "Receiving"? At this point after awaiting `versionExchangePromise`, the version exchange is completed.

And same below for the other case of ReceivingProtocolVersionExchange.

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -750,8 +755,10 @@ export class PortForwardingService extends SshService {
 				this.localForwarders.delete(remotePort);
 			});
 
+			this.raiseReportProgress(Progress.CompletedLocalPortForwarding);
 			return localPort;
 		} else if (remotePort !== 0) {
+			this.raiseReportProgress(Progress.CompletedRemotePortForwarding);
```
jasongin, line 761: The name of this event might be confusing because the distinction here is not about whether the port being forwarded is local or remote, but whether the client is listening on a local socket for the port or not. (For the latter case the app may still open streams for connections to the port.) In both cases the port is being forwarded _from_ the local side _to_ the remote side. The reverse is also possible but is not currently done with dev tunnels.

src/ts/ssh-tcp/services/portForwardingService.ts
```diff
@@ -882,6 +889,7 @@ export class PortForwardingService extends SshService {
 		port: number,
 		cancellation?: CancellationToken,
 	): Promise<SshChannel> {
+		this.raiseReportProgress(Progress.OpeningChannelPortForwarding);
```
jasongin, line 892: I'd suggest OpeningChannelForForwardedPort or OpeningPortForwardingChannel. And similar for Opened below.

src/ts/ssh/progress.ts
```diff
@@ -0,0 +1,39 @@
...
+	CompletedKeyExchange = "CompletedKeyExchange",
+	StartingSessionAuthentication = "StartingSessionAuthentication",
+	CompletedSessionAuthentication = "CompletedSessionAuthentication",
+
+	// Port forwarding messages
+	StartingPortForwarding = "StartingPortForwarding",
+	CompletedLocalPortForwarding = "CompletedLocalPortForwarding",
+	CompletedRemotePortForwarding = "CompletedRemotePortForwarding",
+	OpeningChannelPortForwarding = "OpeningChannelPortForwarding",
+	OpenedChannelPortForwarding = "OpenedChannelPortForwarding",
```
jasongin, line 32: There are already `PortForwarding` and `ChannelOpening` events which have more details than these progress events. Do you think the progress events are still useful? Or should applications listen to progress events for only basic session connection operations and use the other events for port-forwarding and channel opening?

jasongin (approved):

src/ts/ssh/sshSession.ts
jramsay: Thanks for reviewing - even on holidays no less!

I had the same thought, was going to update. Will change that

src/ts/ssh-tcp/services/portForwardingService.ts
jramsay: Will make it CompletedPortForwarding for both. We can make if more specific later if necessary.

src/ts/ssh/progress.ts
jramsay: If there is already more detailed events available, I'll remove them from here. If we find folks are losing the connection thread at that point then we can add them back in later.

src/ts/ssh-tcp/services/portForwardingService.ts
jramsay: actually, removed port forwarding and channel opening events for the time being

jfullerton44 (approved):

PR #89, opened 2023-12-15
Add SshReportProgressEventArgs

jramsay:
- Add SshReportProgressEventArgs that includes a server session number

### DIFF
--- a/src/ts/ssh/sshSession.ts
+++ b/src/ts/ssh/sshSession.ts
@@ -51,6 +51,7 @@ import { Semaphore } from './util/semaphore';
 import { PipeExtensions } from './pipeExtensions';
 import { Queue } from './util/queue';
 import { Progress } from './progress';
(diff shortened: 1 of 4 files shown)


src/ts/ssh/sshSession.ts
```diff
@@ -69,6 +70,12 @@ interface RequestHandler {
 	isCancelled?: boolean;
 }
 
+/**
+ * Allows SSH Server sessions to keep track of the session number for progress
+ * reporting purposes.
+ */
+let serverSessionCounter = 0;
```
IlyaBiryukov, line 77: Why limit it to only server sessions? Can be used for both.

jramsay: Does it make sense in the client context?

jramsay: Thanks! As discussed, could be used for reconnections.

IlyaBiryukov (approved):

PR #90, opened 2023-12-19
[C#] Add ReportProgress infrastructure

jramsay:
[C#] Add ReportProgress infrastructure

### DIFF
--- a/src/cs/Ssh/Events/SshReportProgressEventArgs.cs
+++ b/src/cs/Ssh/Events/SshReportProgressEventArgs.cs
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.DevTunnels.Ssh.Events;
+
+/// <summary>
+/// Event raised to report connection progress.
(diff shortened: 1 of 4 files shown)


src/cs/Ssh/Events/SshReportProgressEventArgs.cs
```diff
@@ -0,0 +1,30 @@
...
+	{
+		Progress = progress.ToString();
+		SessionNumber = sessionNumber;
+	}
+
+	/// <summary>
+	/// Specifies the progress event that is being reported. See <see cref="Progress"/>
+	/// for a description of the different progress events that can be reported.
+	/// </summary>
+	public string Progress { get; }
```
jfullerton44, line 24: Since this is using an enum, can the type be Progress

jfullerton44 (dismissed):

src/cs/Ssh/Events/SshReportProgressEventArgs.cs
jramsay: Thanks, as discussed. I like that better.

jfullerton44 (approved):

PR #96, opened 2024-04-09
Use ESRP to publish to npm

jfullerton44:
Adds esrp publishing to replace manual npm package publishing

### DIFF
--- a/.pipelines/dev-tunnels-ssh-ci.yaml
+++ b/.pipelines/dev-tunnels-ssh-ci.yaml
@@ -194,39 +194,19 @@ jobs:
           nuGetFeedType: external
           publishFeedCredentials: 'dev-tunnels-nuget'
           publishPackageMetadata: true
-
-      - task: Npm@1
-        displayName: Publish ssh npm package to public feed
-        condition: and(succeeded(), eq('${{ parameters.PublishNpmPackage }}', 'true'))
-        continueOnError: True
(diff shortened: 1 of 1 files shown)


.pipelines/dev-tunnels-ssh-ci.yaml
```diff
@@ -194,39 +194,19 @@ jobs:
...
-        inputs:
-          command: custom
-          verbose: false
-          customCommand: publish out/pkg/dev-tunnels-ssh-keys-$(Build.BuildNumber).tgz --access public
-          customRegistry: useNpmrc
-          customEndpoint: 'dev-tunnels-npm'
-
-      - task: Npm@1
-        displayName: Publish ssh-tcp npm package to public feed
-        condition: and(succeeded(), eq('${{ parameters.PublishNpmPackage }}', 'true'))
```
jasongin, line 222: Should the new task have this `condition`?

jasongin: We don't have to make it conditional on the PublishNpmPackage pipeline variable if we just want it to always publish successful builds. But if it's not used then we should remove that pipeline variable.

jfullerton44: I just added the condition to stay consistent with nuget publishing

jasongin (approved):

PR #104, opened 2024-09-24
Add remote port parameter to TCP listener factory

jasongin:
This addresses a bug in the dev tunnels CLI / SDK: it can print the wrong remote port number for an IPv6 client connection endpoint when the original port number was in use.

    > devtunnel connect neat-field-5zxbcmv       
    Connected to tunnel: neat-field-5zxbcmv
    SSH: Forwarding from 127.0.0.1:9001 to host port 9000.
    SSH: Forwarding from [::1]:9001 to host port 9001.

In the example above, the IPv6 port 9001 is actually forwarding to host port 9000, the same as IPv4. The console output there is wrong. (This only happens if port 9000 was in use before the command was run.)

With the `remotePort` parameter added in this change, the dev tunnels SDK  [RetryTcpListenerFactory](https://github.com/microsoft/dev-tunnels/blob/main/cs/src/Connections/RetryTcpListenerFactory.cs) can be updated to use the `remotePort` instead of `localPort` in the message `"Forwarding from {listener.LocalEndpoint} to host port {localPort}."`.

Additionally, this can allow for a future capability for a tunnel client to map host port numbers to different arbitrary client port numbers.

### DIFF
--- a/src/ts/ssh-tcp/tcpListenerFactory.ts
+++ b/src/ts/ssh-tcp/tcpListenerFactory.ts
@@ -37,7 +37,7 @@ export interface TcpListenerFactory {
 		remotePort: number | undefined,
 		localIPAddress: string,
 		localPort: number,
-		canChangePort: boolean,
+		canChangeLocalPort: boolean,
 		cancellation?: CancellationToken,
 	): Promise<net.Server>;
 }
@@ -47,7 +47,7 @@ export class DefaultTcpListenerFactory implements TcpListenerFactory {
 		remotePort: number | undefined,
 		localIPAddress: string,
 		localPort: number,
-		canChangePort: boolean,
+		canChangeLocalPort: boolean,
 		cancellation?: CancellationToken,
 	): Promise<net.Server> {
 		if (!localIPAddress) throw new TypeError('Local IP address is required.');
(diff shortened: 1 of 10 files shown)


src/ts/ssh-tcp/tcpListenerFactory.ts
```diff
@@ -10,11 +10,12 @@ export interface TcpListenerFactory {
 	 * Creates and starts a TCP listener for the specified local network address and port
 	 * number.
 	 *
+	 * @param remotePort The remote port that this local port will connect to (if known).
 	 * @param localIPAddress Local IP address to listen on.
 	 * @param localPort Requested local port to listen on, or 0 to use a random
 	 * available port number.
-	 * @param canChangePort True if the factory is allowed to select a different
-	 * port number than the one that was requested; if false then the factory must either
+	 * @param canChangeLocalPort True if the factory is allowed to select a different
```
IlyaBiryukov, line 17: The parameter name is still `canChangePort` below.

IlyaBiryukov (approved):

src/ts/ssh-tcp/tcpListenerFactory.ts
jasongin: Fixed.

PR #105, opened 2024-09-25
Update to .NET 8 SDK

jasongin:
[.NET 6 end of life is coming up in a couple months](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), and .NET 3.1 has been unsupported for years. So we should stop using those .NET versions. Library projects continue to target `netstandard2.1` and `net48`, so this change does not impact applications that consume the libraries.

 - Update from .NET 6 to .NET 8 SDK in `global.json`.
 - Update from .NET Core 3.1 + .NET 6 to .NET 8 for test and benchmark projects.
 - Update code in `SshAlgorithm.cs` for detecting and loading libssl. .NET 8 changed something that affected the detection. And it added support for libssl v3, though we can still fall back to v1.1 or v1.0.
 - Update PR and CI build steps accordingly.
 - Fix the build step to link libssl on Mac OS, as it was failing on the current Mac OS build agents. Update it to v3, and fix `DYLD_FALLBACK_LIBRARY_PATH` environment variable which doesn't seem to be set in the GH action. (It is supposed to include /usr/local/lib by default, so this shouldn't be a problem for users.)
 - Fix GitHub action permissions for the test-reporter.

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -26,17 +26,21 @@ jobs:
 
     name: Build SSH on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    env:
+      DYLD_FALLBACK_LIBRARY_PATH: $DYLD_FALLBACK_LIBRARY_PATH:/usr/local/lib
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      # On the Mac build machines, libssl 1.1 is installed
-      # but not linked in the path expected by .NET. Fix it now.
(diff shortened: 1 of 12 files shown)


IlyaBiryukov (approved):

jasongin:
I'm still trying to get the crypto tests working on Mac OS again, that might take several more iterations.

jasongin:
> I'm still trying to get the crypto tests working on Mac OS again, that might take several more iterations.

Fixed.

IlyaBiryukov (approved):

PR #115, opened 2025-04-09
Implement keepalive for ssh client or server session in C# SDK.

klvnraju:


src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,35 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
...
 	}
 
+	internal void ImplementKeepAlive(CancellationToken cancellation)
+	{
+		if (Config.KeepAliveTimeoutInSeconds > 0)
+		{
+			keepAliveTimer = new (
+				async (state) =>
+				{
+					if (this.CanAcceptRequests)
```
IlyaBiryukov, line 511: Should we do try-catch? if this lambda is `async void`, an exception here may terminate the process.

src/cs/Ssh/SshSession.cs
```diff
@@ -472,6 +476,9 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
 					this.OnReportProgress(Progress.CompletedProtocolVersionExchange);
 					IsConnected = true;
 				}
+
+				// Implement KeepAlive
+				ImplementKeepAlive(connectCts.Token);
```
IlyaBiryukov, line 481: We shouldn't use `connectCts.Token` here, `connectCts` is disposed right after the session is connected.

src/cs/Ssh/SshSessionConfiguration.cs
```diff
@@ -33,6 +33,7 @@ public class SshSessionConfiguration
 	private bool locked;
 	private bool enableKeyExchangeGuess;
 	private int maxClientAuthenticationAttempts = 5;
+	private int keepAliveTimeoutInSeconds = 0;
```
IlyaBiryukov, line 36: nit: ' = 0' is redundant.

src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,35 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
 		await this.connectCompletionSource.Task.ConfigureAwait(false);
 	}
 
+	internal void ImplementKeepAlive(CancellationToken cancellation)
```
IlyaBiryukov, line 504: Nit: how about `StartKeepAliveTimer` ?

src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,35 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
...
+	{
+		if (Config.KeepAliveTimeoutInSeconds > 0)
+		{
+			keepAliveTimer = new (
+				async (state) =>
+				{
```
IlyaBiryukov, line 510: `&& !IsClosed` ?

src/cs/Ssh/SshSession.cs
klvnraju: Will change it to `this.disposeCancellationSource.Token`

IlyaBiryukov (approved):

jasongin (changes requested):

src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,45 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
...
+							await SendMessageAsync(
+									new SessionRequestMessage() { RequestType = "keepalive@openssh.com", WantReply = true },
+									cancellation)
+								.ConfigureAwait(true);
+
+							// wait for 1 second and check if a response received by then
+							await Task.Delay(KeepAliveResponceTimeoutInMilliSeconds, cancellation).ConfigureAwait(false);
+							if (!keepAliveResponseReceived)
+							{
+								Trace.TraceInformation("Keep alive response not received.");
```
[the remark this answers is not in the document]
klvnraju: Yes, this change is only for logging the keep alive request and response in their host logs. For us to understand if this will fix this issue. I plan on working on the callback that you suggested.

src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,45 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
...
+			keepAliveTimer = new (
+				async (state) =>
+				{
+					try
+					{
+						if (this.CanAcceptRequests && !IsClosed)
+						{
+							this.keepAliveResponseReceived = false;
+							await SendMessageAsync(
+									new SessionRequestMessage() { RequestType = "keepalive@openssh.com", WantReply = true },
```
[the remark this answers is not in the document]
jasongin: Please add a test case for keep-alive during a long-running session request. It would fail with the current code.

src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,45 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
...
+					{
+						Trace.TraceEvent(
+							TraceEventType.Error,
+							SshTraceEventIds.KeepAliveFailed,
+							ex.ToString());
+					}
+				},
+				null,
+				TimeSpan.FromSeconds(Config.KeepAliveTimeoutInSeconds),
+				TimeSpan.FromMilliseconds(Timeout.Infinite));
```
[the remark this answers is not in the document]
klvnraju: Thanks for catching this. I think it was working for me because we are resetting the timer every time we receive a message.

src/cs/Ssh/SshSessionConfiguration.cs
```diff
@@ -518,6 +519,16 @@ public bool EnableKeyExchangeGuess
 		set => EnsureUnlocked().enableKeyExchangeGuess = value;
 	}
 
+	/// <summary>
+	/// Gets or sets the timeout duration for keeping a connection alive in seconds. Ensures the connection is unlocked
+	/// before setting the value.
```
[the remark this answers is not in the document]
klvnraju: Yes. I will be remove that.

src/cs/Ssh/SshSession.cs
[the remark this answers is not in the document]
jasongin: Well `Timeout.InfiniteTimeSpan` is [the same thing](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/Timeout.cs) as `TimeSpan.FromMilliseconds(Timeout.Infinite)`. Just a little more concise.

src/cs/Ssh/SshSession.cs
```diff
@@ -494,6 +501,45 @@ await SendMessageAsync(KeyExchangeInitMessage.None, connectCts.Token)
...
+					{
+						if (this.CanAcceptRequests && !IsClosed)
+						{
+							this.keepAliveResponseReceived = false;
+							await SendMessageAsync(
+									new SessionRequestMessage() { RequestType = "keepalive@openssh.com", WantReply = true },
+									cancellation)
+								.ConfigureAwait(true);
+
+							// wait for 1 second and check if a response received by then
```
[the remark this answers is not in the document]
klvnraju: I changed this logic to see if a response is received before the next keep-alive message is sent. If there is no response by then, we count that as one failure and notify the event with the count. And the SDK user will have the ability to show message/reconnect in case this happens any number of times.

src/cs/Ssh/SshSession.cs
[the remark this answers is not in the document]
klvnraju: Implemented an event that will be triggered every time keep alive response is not delivered and the SDK user has the ability to choose when to reconnect.

src/cs/Ssh/SshSession.cs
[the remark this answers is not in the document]
klvnraju: I have one unit test which checks if the keep-alive request is received and I am still trying to write one which should call the event when response is not received. And I tried adding one test which does not send keep alive for an active connection but that did not work with my old changes (I will try to add it now).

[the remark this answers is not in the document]
jasongin: > I validated that request handlers are not called for keep-alive requests after your suggestion to handle it in HandleMessageAsync.

That wasn't my concern here. The problem is with the keep-alive _response_ messages, which are handled by `HandleMessageAsync(SessionRequestFailureMessage)`. The response (failure and success) messages unfortunately do not include the request type string in the payload, so there is no way to filter those out before calling `InvokeRequestHandler()` which tries to dequeue the next request handler callback. Since there was no request handler added to the `requestHandlers` queue when sending the keep-alive request, the wrong handler may be dequeued when handling the keep-alive response.

Consider the following scenario:
1. Client is idle for some time and therefore sends a keep-alive request. This does not add to the `requestHandlers` queue.
2. Shortly after that (before the keep-alive response is received) the user does some action that causes the client to send a different kind of session request message. This adds a callback to the `requestHandlers` queue.
3. The server replies to the keep-alive request with a failure message.
4. The client dequeues the next callback from the `requestHandlers` queue and incorrectly interprets the keep-alive failure response as the response to the first request.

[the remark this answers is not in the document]
jasongin: > Currently this class assumes that session responses come in the same order the requests were sent (which might be a bug anyway)

I forgot, that's actually fine since [the spec requires it](https://datatracker.ietf.org/doc/html/rfc4254#section-4). But that does not avoid the problematic scenario I described above.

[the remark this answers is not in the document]
klvnraju: Thanks for explaining this to me. I did that change and added unit tests for failure and no keep alive when active test cases.

src/cs/Ssh/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.DevTunnels.Ssh.Test")]
+[assembly: InternalsVisibleTo("Microsoft.DevTunnels.Ssh.Tcp")]
```
jasongin, line 6: I don't think the internals need to be visible to the TCP assembly. It doesn't seem to be used, and anyway I'd rather ensure cross-assembly APIs are public.

As for the tests' use of the internal `KeepAliveRequestReceived` I have a separate suggestion about that, so that can be removed also.

src/cs/Ssh.Tcp/Services/PortForwardingService.cs
```diff
@@ -594,7 +594,7 @@ public async Task WaitForForwardedPortAsync(
 		}
 	}
 
-	protected override async Task OnSessionRequestAsync(
+	protected internal override async Task OnSessionRequestAsync(
```
jasongin, line 597: I don't think this is necessary? Tests are not calling `OnSessionRequestAsync()`.

Same for line 799.

src/cs/Ssh/SshSession.cs
```diff
@@ -171,6 +174,16 @@ internal SshSession(
...
 
+	/// <summary>
+	/// Event is raised when a keep-alive request is sent but no response is received
+	/// </summary>
+	public event EventHandler<SshKeepAliveEventArgs>? KeepAliveRequestFailed;
+
+	/// <summary>
+	/// Internal event used for testing purposes.
+	/// </summary>
+	internal event EventHandler<EventArgs>? KeepAliveRequestReceived;
```
jasongin, line 185: Instead of this internal event just for testing, I'd suggest using trace events:
1. Define a new event ID in `SshTraceEventIds`.
2. When a keep-alive request is received, trace an event with that ID instead of raising `KeepAliveRequestReceived`.
3. In test cases, listen to trace events check and for that ID. See the `TraceChannelData()` test case for an example of this testing approach. You can re-use `TestTraceListener` - make it a top-level class in the test assembly.

Consider also adding trace event IDs for when a keep-alive message is sent and when a keep-alive reply was not received. Those may help with further testing.

jasongin (changes requested):

src/cs/Ssh/AssemblyInfo.cs
klvnraju: The TCP assembly was adding the `TaskExtensions.cs` explicitly and I removed that from Test assembly and did the same change here.

jasongin: I'm not a big fan of `InternalsVisibleTo` because it can lead to unintentional use of internal APIs across assemblies, or laziness in not creating proper public APIs.

src/cs/Ssh/SshSession.cs
klvnraju: Thanks for pointing about the `TestTraceListener`, I was able to test those scenarios with that.

src/cs/Ssh/SshSession.cs
```diff
@@ -1659,7 +1662,7 @@ internal virtual async Task HandleMessageAsync(
 		}
 		else if (message.RequestType == ExtensionRequestTypes.KeepAliveRequest)
 		{
-			this.KeepAliveRequestReceived?.Invoke(this, EventArgs.Empty);
+			Trace.TraceEvent(TraceEventType.Verbose, SshTraceEventIds.KeepAliveRequestReceived);
```
jasongin, line 1665: Add a message string so the trace log will be readable.

jasongin (approved):

jasongin (approved):

PR #118, opened 2025-07-08
Implement keep alive success callback for Codespaces

klvnraju:
We implemented keep alive failure callback and that helped us implement reconnects etc.
This will help in understanding if keep alive messages are being sent. This will help investigate connection issues without downloading the agent logs (These logs are huge and a pain to download)

### DIFF
--- a/src/cs/Ssh/SshSession.cs
+++ b/src/cs/Ssh/SshSession.cs
@@ -181,9 +181,9 @@ public class SshSession : IDisposable
 	public event EventHandler<SshKeepAliveEventArgs>? KeepAliveFailed;
 
 	/// <summary>
-	/// Event is raised when a keep-alive request is sent and a response is received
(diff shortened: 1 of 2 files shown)


src/cs/Ssh/SshSession.cs
```diff
@@ -530,6 +537,10 @@ internal void StartKeepAliveTimer(CancellationToken cancellation)
 							else
 							{
 								keepAliveFailureCount = 0;
+								keepAliveSuccessCount++;
```
JacobRoberts, line 540: I see that `keepAliveSuccessCount` is defined and uninitialized at the top of the file. Will this work as expected if it is still uninitialized when invoked?

JacobRoberts (approved):

IlyaBiryukov (approved):

PR #154, opened 2026-07-23
Fix channel-open cancellation deadlock in ConnectionService

juanpacostaaa:
## Problem

`ConnectionService.HandleMessageAsync` disposed a pending channel's `CancellationTokenRegistration`
while holding `lockObject`. The callback registered in `OpenChannelAsync` also takes `lockObject`,
and `Dispose()` blocks until an in-flight callback completes. So an open cancelling exactly as its
confirmation was processed deadlocked the pump thread, hanging every later `OpenChannelAsync`

## Fix

Dispose the registration **after** releasing the lock. Removing the pending channel under the lock
makes the callback a no-op, so deferring the dispose is safe. Applied to the confirmation and
failure handlers.

## Tests

Added `OpenChannelCancelDuringConfirmationDoesNotDeadlock` and `SessionRemainsUsableAfterOpenCancelRace`,
which force the race via a small internal hook reached by reflection (like `ReconnectTests` does for
`KeyRotationThreshold`).

### DIFF
--- a/test/cs/Ssh.Test/ChannelTests.cs
+++ b/test/cs/Ssh.Test/ChannelTests.cs
@@ -183,11 +183,11 @@ public class ChannelTests : IDisposable
 	/// reflection (as <see cref="ReconnectTests"/> does for KeyRotationThreshold) so the test
 	/// otherwise uses only the public session API. The hook runs on the receive pump thread
 	/// while it holds the ConnectionService lock during open-confirmation processing: it cancels
-	/// the open (whose callback, registered in OpenChannelAsync, also acquires that lock) from
-	/// another thread and then sleeps so the callback is in-flight and blocked on the lock.
-	/// Before the fix, the pump then disposed the cancellation registration while still holding
-	/// the lock, which blocks on the in-flight callback and deadlocks the session; after the fix
-	/// the registration is disposed after the lock is released.
+	/// the open from another thread, using CancellationToken LIFO callback ordering to
+	/// deterministically confirm the internal callback is contending for the held lock before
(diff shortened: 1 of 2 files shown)


adamasmar (dismissed):

test/cs/Ssh.Test/ChannelTests.cs
```diff
@@ -176,6 +177,107 @@ public async Task OpenChannelCancelByAcceptor()
...
+		{
+			// Fire only once, on the first (raced) open confirmation.
+			hookProperty.SetValue(connectionService, null);
+
+			// Cancel from another thread; the open's cancellation callback then contends for
+			// the ConnectionService lock, which the pump currently holds here.
+			Task.Run(() => cancellationSource.Cancel());
+
+			// Give the cancellation callback time to reach and block on the lock.
+			Thread.Sleep(500);
```
klvnraju: This race trigger currently depends on scheduler timing (`Task.Run` + `Thread.Sleep(500)`), which can make the test flaky under CI load. Could we replace the sleep-based window with an explicit synchronization handshake (for example `ManualResetEventSlim` / `TaskCompletionSource`) so we deterministically know the cancellation callback has reached lock contention before continuing?

juanpacostaaa: Good point, thank you! I've fixed this now in the latest commit

klvnraju (dismissed):

juanpacostaaa:
@microsoft-github-policy-service agree company="Microsoft"

plequere-ms (dismissed):

klvnraju (approved):

